PR #11, opened 2025-10-08
Update email example

umeshma:
Update the email example:
- Support checking for noise words: 
  - Load noise words from a noise file.
  - Added default noiseTerms.txt
  - Support compiler callback to ignore noise words
- Refactor creation code
- Fix bugs in tools\test_email.py
- Update .gitIgnore to ignore .vscode files
- Reformatted code to comply with coding standards

### DIFF
--- a/tools/test_email.py
+++ b/tools/test_email.py
@@ -65,7 +65,7 @@ def command(parser: argparse.ArgumentParser):
 
 
 # Just simple test code
-# TODO : Once stable, move creation etc to utool.py
+# TODO : Once stable, move creation etc to query.py
 async def main():
 
     base_path = Path("/data/testChat/knowpro/email/")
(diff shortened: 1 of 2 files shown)


tools/test_email.py
```diff
@@ -65,7 +65,7 @@ def decorator(func: Callable):
 
 
 # Just simple test code
-# TODO : Once stable, move creation etc to query.py
+# TODO : Once stable, move creation etc to utool.py
```
gvanrossum-ms:
```suggestion
# TODO : Once stable, move creation etc to query.py
```

typeagent/emails/email_memory.py
```diff
@@ -19,12 +19,15 @@
...
+    IMessage,
     IMessageCollection,
     ISemanticRefCollection,
     ITermToSemanticRefIndex,
     Term,
 )
-from ..storage.memory import semrefindex
+from ..storage.memory import (
+    semrefindex,
+)
```
gvanrossum-ms:
```suggestion
from ..storage.memory import semrefindex
```

PR #19, opened 2025-10-09
ConversationBase, Email Memory example updates

umeshma:
* ConversationBase - query method updated to permit passing in (optional) options
  * Needed to control topK and other parameters
  * Allow inheritors to customize query behavior
* EmailMemory now uses ConversationBase
  * Reworked entirely. Much cleaner and simpler 
* Email Import:
  * EmailMessage + Meta: also capture message ID and src_Url or file_path
* Improved test_email.py
  * Ingestion: incremental indexing
  * Tracks what messages were already indexed: simplifying testing
  * Useful debug flags for error handling etc
* Bug fixes

### DIFF
--- a/tools/test_email.py
+++ b/tools/test_email.py
@@ -90,8 +90,10 @@ async def main():
 
     if sys.argv[1:2]:
         base_path = Path(sys.argv[1])
-    else:
+    elif os.path.exists("/data"):
         base_path = Path("/data/testChat/knowpro/email/")
+    else:
+        base_path = Path(".")
 
     try:
         base_path.mkdir(parents=True, exist_ok=True)
--- a/typeagent/emails/email_memory.py
+++ b/typeagent/emails/email_memory.py
@@ -173,8 +173,7 @@ async def _add_synonyms_file_as_aliases(
         data: list[dict] = json.load(f)
     if data:
         storage_provider = conversation.settings.storage_provider
-        await storage_provider.begin_transaction()
-        try:
+        async with storage_provider:
             if clean:
                 await aliases.clear()
             for obj in data:
@@ -184,10 +183,6 @@ async def _add_synonyms_file_as_aliases(
                     related_term = Term(text=text.lower())
                     for synonym in synonyms:
                         await aliases.add_related_term(synonym.lower(), related_term)
-            await storage_provider.commit_transaction()
-        except:
-            await storage_provider.rollback_transaction()
-            raise
 
 
 def _add_noise_words_from_file(
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -63,6 +63,12 @@ jobs:
     # - name: Setup runtime (example)
     #   uses: actions/setup-example@v1
 
+    # Force Python 3.12 for this job (uv will pick this up)
+    - name: Set up Python 3.12
+      uses: actions/setup-python@v5
+      with:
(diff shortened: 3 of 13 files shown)


tools/test_email.py
```diff
@@ -39,22 +37,41 @@ def __init__(
...
 
     async def load_conversation(self, db_name: str, create_new: bool = False):
-        await self.conversation.settings.conversation_settings.storage_provider.close()
+        await self.conversation.settings.storage_provider.close()
+        self.db_name = db_name
         self.db_path = self.base_path.joinpath(db_name)
         self.conversation = await load_or_create_email_index(
             str(self.db_path), create_new
         )
+        self.index_log = load_index_log(str(self.db_path), create_new)
```
gvanrossum-ms, line 63: I take it this holds a persistent, mutable mapping of email IDs to booleans? (Or rather a set of email IDs that have been indexed.)

tools/test_email.py
```diff
@@ -39,22 +37,41 @@ def __init__(
...
 
     # Delete the current conversation and re-create it
     async def restart_conversation(self):
-        await self.conversation.settings.conversation_settings.storage_provider.close()
-        self.conversation = await load_or_create_email_index(
-            str(self.db_path), create_new=True
-        )
+        await self.load_conversation(self.db_name, create_new=True)
```
gvanrossum-ms, line 67: So restart is now a synonym for load?

tools/test_email.py
```diff
@@ -91,10 +106,11 @@ async def main():
     print("Email Memory Demo")
     print("Type @help for a list of commands")
 
-    db_path = str(base_path.joinpath("pyEmails.db"))
+    default_db = "gmail.db"  # "pyEmails.db"
```
gvanrossum-ms, line 109: Why not email.db. :-)

tools/test_email.py
```diff
@@ -174,34 +193,46 @@ async def add_messages(context: EmailContext, args: list[str]):
...
+            email_src = email.src_url if email.src_url is not None else ""
+            print_progress(i + 1, None, email.src_url)
+            print()
+            if context.is_indexed(email_id):
+                print(Fore.GREEN + email_src + "[Already indexed]" + Fore.RESET)
+                continue
 
+            try:
+                await conversation.add_messages_with_indexing([email])
+                context.log_indexed(email_id)
```
gvanrossum-ms, line 221: Yeah, so these two aren't transactional (because the second doesn't use SQLite3) and in theory the first could succeed and the second could fail. Not sure what to do about it other than adding an extra column to the messages table for the email_ID. :-(

typeagent/emails/email_import.py
```diff
@@ -14,14 +14,10 @@
 
 def import_emails_from_dir(
```
gvanrossum-ms, line 15: Eventually this file and these functions need to switch to 'ingest'

typeagent/emails/email_memory.py
```diff
@@ -220,7 +158,7 @@ def _is_searchable_term(self, term: str) -> bool:
 
 # Load synonyms from a file and add them as aliases
 async def _add_synonyms_file_as_aliases(
-    conversation: IConversation, file_name: str
+    conversation: ConversationBase, file_name: str, clean: bool
```
gvanrossum-ms, line 161: Rename clean -> clear

PR #22, opened 2025-10-12
Spruce up project a bit

gvanrossum:
- Add docs/README.md and link to it from PyPI and toplevel README.md.
- Tweak AGENTS.md and TADA.md..
- Improve gmail/gmail_dump.py

### DIFF
--- a/TADA.md
+++ b/TADA.md
@@ -40,6 +40,7 @@ Minor:
   - And alternatively (first?) what to put in shell env directly
 - Document test/build/release process
 - Document how to run evaluations (but don't reveal all the data)
+- Document how to use gmail_dump.py (set up a project etc.)
 
 ## Demos
 


gmail/gmail_dump.py
```diff
@@ -1,29 +1,85 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from pathlib import Path
+import argparse
+import time
 from base64 import urlsafe_b64decode as b64d
+from pathlib import Path
+
+from google.oauth2.credentials import Credentials
```
robgruen, line 9: nit: sort order

gmail/gmail_dump.py
```diff
@@ -1,29 +1,85 @@
...
 SCOPES = ["https://www.googleapis.com/auth/gmail.readonly"]
-CREDS = "client_secret.json"
-TOKEN = Path("token.json")
-OUT = Path("mail_dump"); OUT.mkdir(exist_ok=True)
-
-def get_creds():
-    if TOKEN.exists():
-        return Credentials.from_authorized_user_file(TOKEN, SCOPES)
-    flow = InstalledAppFlow.from_client_secrets_file(CREDS, SCOPES)
+OUT = Path("mail_dump")
```
robgruen, line 14: If this isn't in the .gitignore then people might accidentally add files to the repo.  I only call this out since you were direct in your statements above about adding to the repo.  If it's not in the .gitignore, maybe change this to _HOME_DIR/mail_dump_ or something like that.

README.md
```diff
@@ -12,7 +12,7 @@ This is an in-progress project aiming at a Pythonic translation of
 
 ### Documentation
 
-(To be written. Sorry.)
+Found in the [docs directory](docs/README.md)
```
robgruen, line 20: Maybe add a statement about how e-mail will be send to the LLM for indexing and as such people shouldn't use this tool to export sensitive data.

gmail/gmail_dump.py
gvanrossum: The default location when running the script from gmail/ is in the .gitignore.

Also, I think people who blindly add every new file to the repo should not be using git. :-)

PR #23, opened 2025-10-12
Update field_helpers.py

supadrupa:


gvanrossum:
Thanks for catching this. I'm not sure if I'm set up to take 3rd party PRs yet (did you encounter a CLA screen?) so I will apply your change in one of my own PRs (with acknowledgement using Co-authored-by: supadrupa) unless I find out soon that we're set up right.

gvanrossum:
Never mind. I see a note that all CLA requirements are met. I will merge this now.

PR #27, opened 2025-10-16
Email Knowledge Extraction bug

umeshma:
Fixed an issue in how email addresses were getting turned into actions.

### DIFF
--- a/typeagent/emails/email_message.py
+++ b/typeagent/emails/email_message.py
@@ -113,14 +113,25 @@ class EmailMessageMeta(IKnowledgeSource, IMessageMetadata):
     def _createActions(
         self, verb: str, sender: str, recipient: str
     ) -> list[kplib.Action]:
-        display_name, address = parseaddr(recipient)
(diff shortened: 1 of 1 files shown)


typeagent/emails/email_message.py
```diff
@@ -113,15 +113,26 @@ def _email_address_to_entities(
...
+        self, actions: list[kplib.Action], verb: str, sender: str, recipient: str
+    ) -> None:
+        recipient_display_name, recipient_address = parseaddr(recipient)
+        if recipient_display_name:
+            actions.append(self._createAction(verb, sender, recipient_display_name))
+
+        if recipient_address:
+            actions.append(self._createAction(verb, sender, recipient_address))
+
     def _createAction(
```
gvanrossum-ms, line 136: FWIW, this function is now no longer used.

gvanrossum-ms: Oops, never mind.

PR #32, opened 2025-10-19
Update env-vars.md with new variables for local hosted models

wirthual:
Added OpenAI variables which are required to run the example with locally hosted models.

Example got tested with locally hosted embedding model:
```bash
docker run -it -p 7997:7997  michaelf34/infinity:latest  v2  --model-id mixedbread-ai/mxbai-embed-large-v1 --device cpu --engine optimum
```

Together with a chat completion model:
```bash
ollama run llama3.2:1b
```

I was able to execute the `Getting Started` guide with the following command:
```bash
OPENAI_BASE_URL='http://localhost:7997' OPENAI_ENDPOINT='http://localhost:11434/v1/chat/completions' OPENAI_API_KEY='does-not-matter' OPENAI_MODEL="llama3.2:1b" uv run python code.py
```

Which resulted in the following output:
```bash
0.044s -- Using OpenAI
Indexing 3 messages...
Indexed 3 messages.
Got 15 semantic refs.
```

The used model uses embedding dimensions of 1024, which I adapted by changing `DEFAULT_EMBEDDING_SIZE`

### DIFF
--- a/docs/env-vars.md
+++ b/docs/env-vars.md
@@ -18,8 +18,8 @@ The (public) OpenAI environment variables include:
 - `OPENAI_MODEL`: An environment variable introduced by
   [TypeChat](https://microsoft.github.io/TypeChat/docs/examples/)
   indicating the model to use (e.g.`gpt-4o`).
-- `OPENAI_BASE_URL`: The URL for an openai compatible embedding server, e.g. [Infinity](https://github.com/michaelfeil/infinity). With this option `OPENAI_API_KEY` also needs to be set, but can be any value.
-- `OPENAI_ENDPOINT`: The URL for an server compatible with the OpenAI Chat Completions API. Make sure the `OPENAI_MODEL` variable matches with the deployed model name. E.g. 'llama:3.2:1b'
+- `OPENAI_BASE_URL`: **Optional:** The URL for an OpenAI-compatible embedding server, e.g. [Infinity](https://github.com/michaelfeil/infinity). With this option `OPENAI_API_KEY` also needs to be set, but can be any value.
+- `OPENAI_ENDPOINT`: **Optional:** The URL for an server compatible with the OpenAI Chat Completions API. Make sure the `OPENAI_MODEL` variable matches with the deployed model name, e.g. 'llama:3.2:1b'
 
 ## Azure OpenAI environment variables
 


docs/env-vars.md
```diff
@@ -12,12 +12,13 @@ Typeagent currently supports two families of environment variables:
 ## OPENAI environment variables
 
 The (public) OpenAI environment variables include:
-
```
gvanrossum, line 15: Can I get this blank line back? I like the extra bit of whitespace in the rendered page.

docs/env-vars.md
```diff
@@ -12,12 +12,13 @@ Typeagent currently supports two families of environment variables:
...
 - `OPENAI_API_KEY`: Your secret API key that you get from the
   [OpenAI dashboard](https://platform.openai.com/api-keys).
 - `OPENAI_MODEL`: An environment variable introduced by
   [TypeChat](https://microsoft.github.io/TypeChat/docs/examples/)
   indicating the model to use (e.g.`gpt-4o`).
+- `OPENAI_BASE_URL`: The URL for an openai compatible embedding server, e.g. [Infinity](https://github.com/michaelfeil/infinity). With this option `OPENAI_API_KEY` also needs to be set, but can be any value.
+- `OPENAI_ENDPOINT`: The URL for an server compatible with the OpenAI Chat Completions API. Make sure the `OPENAI_MODEL` variable matches with the deployed model name. E.g. 'llama:3.2:1b'
```
gvanrossum:
```suggestion
- `OPENAI_BASE_URL`: **Optional:** The URL for an OpenAI-compatible embedding server, e.g. [Infinity](https://github.com/michaelfeil/infinity). With this option `OPENAI_API_KEY` also needs to be set, but can be any value.
- `OPENAI_ENDPOINT`: **Optional:** The URL for an server compatible with the OpenAI Chat Completions API. Make sure the `OPENAI_MODEL` variable matches with the deployed model name, e.g. 'llama:3.2:1b'
```

docs/env-vars.md
wirthual: Done.

gvanrossum-ms:
Thanks for your contribution! Really cool you got it to work with a local model.

Regarding the embedding width, I *think* the correct way currently is to update the dict `model_to_embedding_size_and_envvar` in typeagent/aitools/embeddings.py with an appropriate entry (you could patch it in from the outside, that variable is not private) and then use code like this:

```py
from typeagent.aitools.embeddings import AsyncEmbeddingModel
from typeagent.knowpro.convsettings import ConversationSettings

async def main():
    embedding_model = AsyncEmbeddingModel(embedding_size=1024, model_name="your-name")
    settings = ConversationSettings(model=embedding_model)
    settings.semantic_ref_index_settings.auto_extract_knowledge = True      # DON'T FORGET THIS
    conversation = await create_conversation("demo.db", TranscriptMessage, settings=settings)
    # ... original code ...
```

You could argue that it should be possible to construct an `AsyncEmbeddingModel` instance that doesn't correspond to a table entry. It looks like there's already an argument `endpoint_var` but no logic that uses it. If you're interested I encourage you to submit a PR.

PR #33, opened 2025-10-19
Push 0.3.0 release; add slides from PyBay talk

gvanrossum-ms:


gvanrossum (approved):

jf---:
hi @gvanrossum-ms, I've tried to check out the shared slide deck, it seems to me that an organization account that has permission is required to view the slides? Seems I can't check out the slides with the current protection. Thanks for an amazing talk!

gvanrossum:
Sorry about that. The version in the 'release' branch works.
https://github.com/microsoft/typeagent-py/blob/release/docs/StructuredRagPyBay25.pptx

gvanrossum:
Now also fixed in main.

PR #49, opened 2025-10-23
add retry functionality to extract_knowledge_from_text

wirthual:
Should there be a delay in between the calls? With a potential backoff?

### DIFF
--- a/test/test_knowledge.py
+++ b/test/test_knowledge.py
@@ -15,6 +15,7 @@ from typeagent.knowpro.knowledge import (
 from typeagent.knowpro import convknowledge, kplib
 
 from fixtures import really_needs_auth
+from unittest.mock import AsyncMock
 
 
 class MockKnowledgeExtractor:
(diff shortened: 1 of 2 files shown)


gvanrossum-ms:
I'm sorry, this currently adds retries to knowledge extraction. However, that uses typechat, which already implements its own retry logic -- so we don't need to add our own.

I was thinking of adding it to the guts of AsyncEmbeddingModel, in `get_embeddings_nocache()` (the non-test branch towards the end). But we should first investigate whether `AsyncOpenAI.embeddings.create` already does or supports retries. (It has a timeout argument, but nothing to control retries that I can see.)

Assuming we have to do our own retries, the next thing to try is if its return value gives you access to the raw HTTP headers. There should be a header `Retry-After:`.

Maybe you can copy some logic from [TypeAgent](https://github.com/microsoft/TypeAgent/blob/main/ts/packages/aiclient/src/restClient.ts) -- search for `Retry-After`.

gvanrossum:
If no retry-after header is found we should definitely use exponential backoff, e.g. first retry after 1 second, second after 2s, then 4s, 8s etc.

wirthual:
Understood.  Looks like `AsyncOpenAI` and `AsyncAzureOpenAI` support specifying the retries at [initialization](https://github.com/openai/openai-python/blob/dff16b5e9964bf85157eb41181255ed5dde8dda4/src/openai/_client.py#L460).

gvanrossum:
Oh in that case we don't need to do anything. The default value is 2 retries, which brings the total number of tries to 3(*), which is exactly what I'd like to use.

_____
(*) The code is pretty convoluted but I found code that loops over `range(max_retries + 1)` so clearly it tries three times if max_retries is set to two.

PR #59, opened 2025-10-29
Misc changes

gvanrossum-ms:
Latest to earliest commit:

* cd4e870 Mark two more tests as really_needs_auth
* 08f928c Fix pyright 1.1.406, since 407 has a regression (https://github.com/microsoft/pyright/issues/11060)
* 10b8b69 Reformat vectorbase.py
* 969d130 Cleanup deps
* a2a3ff3 Default auto_extract_knowledge to True
* 707e8b8 Tweak contents and initialization of TextEmbeddingIndexSettings
     - Remove retry_max_attempts, retry_delay
     - Add max_retries
     - More sensible initializations
     - Pass max_retries to AsyncEmbeddingModel
     - Default batch_size to 10
* a53e873 ingest_vtt.py: take batch size from settings
* d157257 Fix some embeddings issues
     - add max_retries arg, passed to openai.AsyncOpenAI
     - Fix typos in model_to_embedding_size_and_envvar keys (missing -3- in names)
* be77839 Upgrade dependencies, including pyright
* fcce50f Update TODO list
* 8999559 Use black.Mode instead of obsolete black.FileMode

### DIFF
--- a/test/test_factory.py
+++ b/test/test_factory.py
@@ -10,6 +10,8 @@ from typeagent.aitools.embeddings import AsyncEmbeddingModel, TEST_MODEL_NAME
 from typeagent.knowpro.convsettings import ConversationSettings
 from typeagent.transcripts.transcript import TranscriptMessage, TranscriptMessageMeta
 
+from fixtures import really_needs_auth
+
 
 @pytest.mark.asyncio
 async def test_create_conversation_minimal():
@@ -51,7 +53,7 @@ async def test_create_conversation_with_tags():
 
 
 @pytest.mark.asyncio
-async def test_create_conversation_and_add_messages():
+async def test_create_conversation_and_add_messages(really_needs_auth):
     """Test the complete workflow: create conversation and add messages."""
     # 1. Create empty conversation
     test_model = AsyncEmbeddingModel(model_name=TEST_MODEL_NAME)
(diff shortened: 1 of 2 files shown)


gvanrossum-ms:
Whoops, there are type checking errors that I need to fix later.

gvanrossum (approved):

typeagent/aitools/vectorbase.py
```diff
@@ -18,27 +19,34 @@ class ScoredInt:
...
         self,
         embedding_model: AsyncEmbeddingModel | None = None,
         embedding_size: int | None = None,
         min_score: float | None = None,
         max_matches: int | None = None,
+        batch_size: int | None = None,
+        max_retries: int | None = None,
     ):
-        self.embedding_model = embedding_model or AsyncEmbeddingModel(embedding_size)
+        self.min_score = min_score if min_score is not None else 0.85
```
robgruen, line 37: Maybe consider adding a check to ensure min_score falls between 0 & 1 inclusively in addition to None

gvanrossum-ms: Thanks -- though I think nothing will go really wrong if this is out of range. I have range checks for `max_matches` and `batch_size` because `<= 0` there would cause actual problems (e.g. batch size 0 will cause an infinite loop) and I like that passing `0` means "use the default". Each parameter is just a little unique.

I'll merge as-is.

Issue #45, opened 2025-10-22
Let's make the email demo really good


gvanrossum-ms:
So good that I'll actually want to use it for searching my own email instead of GMail's own search (which is okay but not amazing, and recent tweaks just make it more chaotic).

But not so good that it becomes a service that everyone can just subscribe to -- I don't want to deal with running a service that might go down, etc.

gvanrossum-ms:
@umeshma

bmerkle:
small idea: for the email data so that we do not have to use gmail accounts for demos...

we could use public available datasets and download them e.g. from huggingface or kaggle,
or download them a priori and store it in the demo git repo. (so if would be selfcontained)

Using the Enron Email Dataset contains ~500,000 messages from ~150 users at Enron.
`from datasets import load_dataset`
`ds = load_dataset("enron_email")`
`print(ds["train"][0])`


Available also at [Kaggle](https://www.kaggle.com/wcukierski/enron-email-dataset)  or [Carnegie Mellon](https://www.cs.cmu.edu/~enron/)
.

gvanrossum:
Good idea -- though I currently expect that ingesting 500,000 messages could take weeks to index.

I was focusing on my own emails because that's my actual use case. :-)

bmerkle:
yes, lets take only a fraction of them :) or maybe i find a smaller public set
if its ok, i could create a PR for it. 

the approach would be similar to sklearn
`from sklearn.datasets import fetch_20newsgroups`

the old times when there were newsgroups :) transferred with uucp and cnews

gvanrossum:
I like including some of that in testdata or having a test that downloads and caches it.

But beware that both those corpora are over 20 years old, and the way people write and mark up their emails have changed. In part because the tools are imposing new conventions for quoting -- what Outlook and GMail do is quite different from what we did around 2000.

bmerkle:
@gvanrossum small question: are there specific points in which direction we should improve the demo ?
We have discussed about the (mail) data, but to do concrete steps, i would need a few concrete topics/plan

I also looked at TODO.md and would be open to take some tasks: maybe first start with a easy task from 
- Code Quality & Cleanup 
- Documentation
- Development Infra
and via this I get also more insight into the codebase

gvanrossum:
Here are some things wrong with the current email parsing.

[...]

PR #60, opened 2025-10-29
Add batching to embedding model

### LINKED ISSUE #52: Limit embeddings batch size
The fundamental call to get embeddings, `AsyncEmbeddingModel.get_embeddings_nocache()`, does not protect itself against a request containing more text than the server endpoint supports.

It behooves us to use OpenAI's `tiktoken` library to accurately count the tokens (or perhaps to use character count divided by 3 or 4 as an approximation) and if the token count is over 8k (the typical limit) split the batch up into multiple batches each under that limit.

We might even parallelize processing of those batches (using e.g. `asyncio.gather()`), though I've not had much success in other places using concurrency, so I'd do that as an optional separate task.

wirthual:
PR to add batching to embedding model.

For models with tiktoken tokenizer, encode input and batch after this criteria:
 - not a single input larger than `4096` token (doc says 8192 but it fails, 8191 works)
 - the list of inputs not larger than `2048`
 - overall tokens not larger than `300_000`

If a single input is longer than `4096` tokens, the input will be split into chunks and the final embedding is the average over the chunk embeddings (could be also weighted based on the length of the chunks)

Tokenized input is send to model.

Models which do not use tiktoken tokenizer continue to operate on strings.

Docs: https://platform.openai.com/docs/api-reference/embeddings/create#embeddings_create-input

Fixes #52.

### DIFF
--- a/test/test_embeddings.py
+++ b/test/test_embeddings.py
@@ -133,7 +133,7 @@ async def test_refresh_auth(
 
 
 @pytest.mark.asyncio
-async def test_set_endpoint(monkeypatch):
+async def test_set_endpoint(monkeypatch: MonkeyPatch):
     """Test creating of model with custom endpoint."""
 
     monkeypatch.setenv("AZURE_OPENAI_API_KEY", "does-not-matter")
@@ -203,7 +203,7 @@ async def test_set_endpoint(monkeypatch):
 
 @pytest.mark.asyncio
 async def test_embeddings_batching_tiktoken(
-    fake_embeddings_tiktoken: FakeEmbeddings, monkeypatch
+    fake_embeddings_tiktoken: FakeEmbeddings, monkeypatch: MonkeyPatch
 ):
     monkeypatch.setenv("OPENAI_API_KEY", "test_key")
 


gvanrossum:
Ah, now I understand what you meant by averaging embeddings for long inputs. It'll take me a while to review this, it's the most substantial contribution ever! :-)

gvanrossum:
Also could you work on the merge conflicts? Unfortunately I just did some work on the embedding code myself...

typeagent/aitools/utils.py
```diff
@@ -299,3 +304,8 @@ def make_agent[T](cls: type[T]) -> Agent[None, T]:
         )
 
     return Agent(model, output_type=Wrapper(cls, strict=True), retries=3)
+
+
+def chunk_input(data: list[T] | str, chunk_size: int) -> Iterator[list[T] | str]:
```
gvanrossum: This would be a good place to use `@overload` to indicate that if the input is a list[T], so is the output, and ditto str->str.

Separately, maybe this should accumulate values in a list rather than yielding them, since the caller immediately casts to a list, and the many context switches between caller and generator just slow it down.

typeagent/aitools/utils.py
```diff
@@ -15,7 +15,12 @@
 import dotenv
 import typechat
 
+from itertools import batched
+
 from pydantic_ai import Agent
+from typing import Iterable, Iterator, List, TypeVar
+
+T = TypeVar("T")
```
gvanrossum: We don't use TypeVar any more. We use the new (PEP 695) notation, where you write `def func[T](args): ...`

We require Python 3.12 or higher, where this is supported.

typeagent/aitools/utils.py
```diff
@@ -15,7 +15,12 @@
 import dotenv
 import typechat
 
+from itertools import batched
+
 from pydantic_ai import Agent
+from typing import Iterable, Iterator, List, TypeVar
```
gvanrossum: Don't use List or TypeVar

typeagent/aitools/utils.py
```diff
@@ -15,7 +15,12 @@
 import dotenv
 import typechat
 
+from itertools import batched
```
gvanrossum: Looks unused?

typeagent/aitools/embeddings.py
```diff
@@ -3,10 +3,15 @@
 
 import asyncio
 import os
+import re
 
 import numpy as np
 from numpy.typing import NDArray
 from openai import AsyncOpenAI, AsyncAzureOpenAI, DEFAULT_MAX_RETRIES, OpenAIError
+import tiktoken
+from tiktoken import model
```
gvanrossum: Can we rename that to `import model as tiktoken_model`? `model` feels too close to a likely local variable. Or maybe just spell it as `tiktoken.model`.

typeagent/aitools/embeddings.py
```diff
@@ -37,6 +47,7 @@ class AsyncEmbeddingModel:
     async_client: AsyncOpenAI | None
     azure_endpoint: str
     azure_api_version: str
+    encoding_name: str | None
```
gvanrossum: Looks like you don't ever set this variable. Instead, you set several others, below on lines 114-120, that you might want to declare here.

typeagent/aitools/embeddings.py
```diff
@@ -188,18 +209,81 @@ def hashish(s: str) -> int:
...
-            assert len(data) == len(input), (len(data), "!=", len(input))
-            return np.array([d.embedding for d in data], dtype=np.float32)
+
+            chunked_groups = {}
+            chunked_input = []
+            chunked_sizes = []
+            for input_idx, entity in enumerate(input):
+                entity_to_embed = entity
+                if self.encoding:
+                    entity_to_embed = self.encoding.encode(entity)
```
gvanrossum: I find it hard to think about the code because `entity_to_embed` could be either a `str` or a `list[int]` (which `encode()` returns).

Moreover I think that in the case that it's a `list[int]`, the rest of the code is wrong -- we end up sending lists of integers to the embedding endpoint, which AFAIK only expects lists of strings.

typeagent/aitools/utils.py
```diff
@@ -299,3 +304,8 @@ def make_agent[T](cls: type[T]) -> Agent[None, T]:
         )
 
     return Agent(model, output_type=Wrapper(cls, strict=True), retries=3)
+
+
+def chunk_input(data: list[T] | str, chunk_size: int) -> Iterator[list[T] | str]:
+    for i in range(0, len(data), chunk_size):
+        yield data[i : i + chunk_size]
```
gvanrossum: When data is a `str`, this is going to split things in the middle of words, which is definitely causing the embedding to misunderstand the boundaries. Even if you split between words it's debatable, since you might be splitting in the middle of a key phrase or concept (e.g. "pushing up the daisies").

typeagent/aitools/embeddings.py
```diff
@@ -188,18 +209,81 @@ def hashish(s: str) -> int:
...
-                    input=input,
-                    model=self.model_name,
-                    encoding_format="float",
-                    **extra_args,
-                )
-            ).data
-            assert len(data) == len(input), (len(data), "!=", len(input))
-            return np.array([d.embedding for d in data], dtype=np.float32)
+
+            chunked_groups = {}
```
gvanrossum: Here starts a very long block of code that I think would gain a lot of clarity if you refactored it into a few helper functions/methods.

gvanrossum (changes requested):
Thanks! This is a good start.

I haven't reviewed everything yet, but I believe I've found a pretty big flaw (lists of ints) in the chunking code.

I also found a bunch of nits. And I think the code clarity would improve if everything was typed (something I try to enforce in my own code too).

gvanrossum:
@wirthual, I have another thought. Two actually.

First, the *only* place where we ever request embeddings of *long* inputs (controlled by the user) is when adding an index entry for a very long message to `MessageTextIndex`. However! We don't actually look up anything by embedding in that table. The embeddings are just used for a non-existent fallback if regular querying doesn't work, OR when explicitly using classic RAG to compare its effectiveness to structured RAG. But we don't have such a fallback and we don't do that comparison. (They were part of the original TypeScript code base and were somewhat mindlessly translated.)

So I don't think it's worth to complicate the code at all to deal with over-long single embedding input strings, except in order to make sure the entire batch isn't rejected by the embedding service. For this purpose simply truncating over-long input strings would seem good enough, and you won't need `chunked_groups` and any logic around it at all.

OTOH (this is the second thought) if you really, really, really want this embedding API completely rock solid for other use cases, can you research a bit what a typical classic RAG implementation does for over-long input strings?

wirthual:
Hi,

Based on your description it makes sense to just make sure we do not run into the limits of the embedding service by truncating the inputs and limit the batch size. I can simplify this PR to account for that.

Generally splitting on sentence/section level rather than fixed size seems already a better approach. For example the implementation of [RecursiveCharacterTextSplitter](https://reference.langchain.com/python/langchain_text_splitters/?h=textspl#langchain_text_splitters.RecursiveCharacterTextSplitter) does this, where it recursively tries to split the text at certain characters, e.g. `"\n\n", "\n", " ", ""` to prevent text being cut off in the middle of an idea.


There are also more sophisticated [methods](https://github.com/FullStackRetrieval-com/RetrievalTutorials/blob/main/tutorials/LevelsOfTextSplitting/5_Levels_Of_Text_Splitting.ipynb) which take the document type and structure or the semantics of the text into account e.g. with the help of an LLM.

typeagent/aitools/embeddings.py
wirthual: The openAI endpoint also allows sending the tokenized inputs to get the embeddings. So this way it just skips the decoding part. However I understand this is a bit confusing and it might be clearer if the token chunks get decoded again and sent then.

gvanrossum:
@wirthual 
> Based on your description it makes sense to just make sure we do not run into the limits of the embedding service by truncating the inputs and limit the batch size. I can simplify this PR to account for that.

Sounds good!

> Generally splitting on sentence/section level rather than fixed size seems already a better approach. For example the implementation of [RecursiveCharacterTextSplitter](https://reference.langchain.com/python/langchain_text_splitters/?h=textspl#langchain_text_splitters.RecursiveCharacterTextSplitter) does this, where it recursively tries to split the text at certain characters, e.g. `"\n\n", "\n", " ", ""` to prevent text being cut off in the middle of an idea.

Yeah, this should be left up to the process parsing the original input text into message chunks.

> There are also more sophisticated [methods](https://github.com/FullStackRetrieval-com/RetrievalTutorials/blob/main/tutorials/LevelsOfTextSplitting/5_Levels_Of_Text_Splitting.ipynb) which take the document type and structure or the semantics of the text into account e.g. with the help of an LLM.

Also thank you for reminding me that a lot of this stuff has already been done or at least been thought about. (Have you heard of something that can split email messages? That might help with #45...)

typeagent/aitools/utils.py
wirthual: Not used anymore

typeagent/aitools/utils.py
wirthual: outdated

typeagent/aitools/utils.py
wirthual: removed

typeagent/aitools/embeddings.py
wirthual: changed to `from tiktoken import model as tiktoken_model
`

typeagent/aitools/embeddings.py
wirthual: Adressed.

typeagent/aitools/utils.py
wirthual: Using truncation now.

typeagent/aitools/embeddings.py
wirthual: Updated version does not use `chunked_groups` anymore.

typeagent/aitools/embeddings.py
gvanrossum: Oh, if sending tokens is supported, and we are getting tokens from tiktoken anyways, by all means send the tokens! I just didn't realize the API did that. (I looked in the docs but apparently in the wrong place -- do you have a deep link?)

test/fixtures.py
```diff
@@ -39,6 +39,10 @@
 from typeagent.storage.memory import MemoryStorageProvider
 from typeagent.storage import SqliteStorageProvider
 
+from openai.types.create_embedding_response import CreateEmbeddingResponse, Usage
+from openai.types.embedding import Embedding
+import tiktoken
+
```
gvanrossum: Please move this block before the first typeagent import (line 12).

I like the following order for imports:
1. stdlib
2. 3rd party deps
3. deps from this project

And alphabetized by fully qualified module name within each section.

typeagent/aitools/embeddings.py
```diff
@@ -210,80 +212,39 @@ def hashish(s: str) -> int:
             return result
         else:
```
gvanrossum: Don't need this blank line.

test/test_embeddings.py
```diff
@@ -197,3 +197,76 @@ async def test_set_endpoint(monkeypatch):
     # Not even when default model name specified explicitly
     with pytest.raises(ValueError):
         AsyncEmbeddingModel(1024, "text-embedding-ada-002")
+
+
+@pytest.mark.asyncio
+async def test_embeddings_batching_tiktoken(
+    fake_embeddings_tiktoken: FakeEmbeddings, monkeypatch
```
gvanrossum: Curious if the *type* of `monkeypatch` is importable from pytest, so we can make this code more typed?

typeagent/aitools/embeddings.py
wirthual: Yes, I got it from [here](https://github.com/openai/openai-python/blob/6574bcd612771186995074846253caa0ff1ba517/src/openai/resources/embeddings.py#L66)
 > Input text to embed, encoded as a string or array of tokens.

typeagent/aitools/utils.py
wirthual: Removed.

test/test_embeddings.py
```diff
@@ -197,3 +197,76 @@ async def test_set_endpoint(monkeypatch):
     # Not even when default model name specified explicitly
     with pytest.raises(ValueError):
         AsyncEmbeddingModel(1024, "text-embedding-ada-002")
+
```
gvanrossum, line 202: I think you're missing tests for `MAX_TOKENS_PER_BATCH `? I see a test for `MAX_CHARS_PER_BATCH` but not the token-based one?

gvanrossum (changes requested):
Thanks -- great progress. I have a few nits only.

test/fixtures.py
wirthual: Moved the block

typeagent/aitools/embeddings.py
wirthual: Removed

test/test_embeddings.py
wirthual: imported MonkeyPatch from pytest and added typehint

test/test_embeddings.py
wirthual: Good catch. Added a test for `MAX_TOKENS_PER_BATCH`

gvanrossum (approved):
LGTM.

PR #71, opened 2025-11-07
Reinvent conversation metadata API and implementation

gvanrossum:


typeagent/storage/sqlite/provider.py
```diff
@@ -300,62 +333,141 @@ async def deserialize(self, data: dict) -> None:
...
-                cursor.execute(
-                    f"UPDATE ConversationMetadata SET {', '.join(updates)}",
-                    params,
-                )
-        else:
-            # Insert new with default values
+        if not cursor.fetchone():
+            # Insert default values if no metadata exists
             name_tag = f"conversation_{self.conversation_id}"
             schema_version = "1.0"
```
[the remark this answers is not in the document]
gvanrossum: No, it should be str(CONVERSATION_SCHEMA_VERSION). I'll do it in a separate PR, it requires an extra import.

typeagent/storage/sqlite/provider.py
```diff
@@ -300,62 +333,141 @@ async def deserialize(self, data: dict) -> None:
...
+        Args:
+            **kwds: Metadata keys and values where:
+                - str value: Sets a single key-value pair (replaces existing)
+                - list[str] value: Sets multiple values for the same key
+                - None value: Deletes all rows for the given key
+
+        Example:
+            provider.set_conversation_metadata(
+                name_tag="my_conversation",
+                schema_version="0.1",
```
robgruen: nit: Update the comment to reflect schema_version is now just an int.

typeagent/storage/sqlite/provider.py
```diff
@@ -300,62 +333,141 @@ async def deserialize(self, data: dict) -> None:
...
+                return datetime.fromisoformat(value_str)
+            except ValueError:
+                # Fallback for other formats
+                return datetime.now(timezone.utc)
+
+        # Extract standard fields (single values expected)
+        name_tag = get_single("name_tag")
+        schema_version_str = get_single("schema_version", "1")
+        try:
+            schema_version = int(schema_version_str)
```
robgruen, line 378: I don't see where action is being taken against this value (i.e. if it's not the expected version or incompatible with the current version, etc.).  If it is please ignore this comment. 

Maybe just add a check to make sure that this version is >= the version of the schema being loaded?

gvanrossum: It's passed into ConversationMetadata() at the end of the function (L403).

I intentionally don't check against the schema version, because there's only one. Once we have more than one version there will be code that uses it to know what to do (e.g. migrate the db).

But you're right that rejecting a db if its schema has a different version is also a good idea. I think I'll put that in a different PR though (this one is already a bit overweight :-).

robgruen (approved):

PR #74, opened 2025-11-08
Pass name, tags, extras through create_conversation()

gvanrossum:
Also update ingest_vtt.py to use the new API.

### DIFF
--- a/tools/ingest_vtt.py
+++ b/tools/ingest_vtt.py
@@ -25,6 +25,12 @@ import webvtt
 
 from typeagent.aitools import utils
 from typeagent.aitools.embeddings import AsyncEmbeddingModel
+from typeagent.knowpro.convsettings import ConversationSettings
+from typeagent.knowpro.interfaces import ConversationMetadata
+from typeagent.knowpro.universal_message import (
+    UNIX_EPOCH,
+    format_timestamp_utc,
+)
 from typeagent.storage.utils import create_storage_provider
 from typeagent.transcripts.transcript_ingest import (
     get_transcript_duration,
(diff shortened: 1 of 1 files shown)


test/test_conversation_metadata.py
```diff
@@ -498,29 +494,29 @@ async def test_very_long_conversation_id(
...
         self, temp_db_path: str, embedding_model: AsyncEmbeddingModel
     ):
-        """Test conversation metadata with Unicode conversation ID."""
+        """Test conversation metadata with Unicode name_tag."""
         embedding_settings = TextEmbeddingIndexSettings(embedding_model)
         message_text_settings = MessageTextIndexSettings(embedding_settings)
         related_terms_settings = RelatedTermIndexSettings(embedding_settings)
 
-        unicode_id = "конверсация_🚀_测试"  # Mixed Unicode
+        unicode_name = "conversation_конверсация_🚀_测试"  # Mixed Unicode
```
robgruen, line 511: Nice!

typeagent/storage/sqlite/provider.py
```diff
@@ -348,20 +359,19 @@ def get_conversation_metadata(self) -> ConversationMetadata | None:
...
+        def get_single(key: str) -> str | None:
+            values = metadata_dict.get(key)
+            if values is None:
+                return None
             if len(values) > 1:
                 raise ValueError(
                     f"Expected single value for key '{key}', got {len(values)}"
                 )
-            return values[0] if values else default
+            return values[0]
```
robgruen, line 371: Could values ever be empty here?  There's no check for len(values) == 0

gvanrossum: The way metadata_dict is maintained (see some lines above), values is never `[]`.

PR #83, opened 2025-11-12
use concurrency for create_knowledge_extractor

KRRT7:
Screenshot 2025-11-12 at 5 13 07 AM

[codeflash](https://www.codeflash.ai) originally found this optimization as seen above, but we had a bug which prevented us from automatically opening the PR, so I'm doing it manually

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -17,8 +17,7 @@ Gradually move work items from here to repo Issues.
 ## Storage & Persistence
 
 ### P0 - Critical
-- **[P0, large]** Fix all bugs related to ordinals/ids relying on starting at 0 and no gaps
-- **[P0, small]** Need embedding size and embedding name in metadata
+- **[P0, large]** Fix all bugs related to ordinals/ids relying on starting at 0 and no gaps (prepare for deletions)
 
 ### P1 - High Priority
 - **[P1, medium]** Scrutinize sqlite/reltermsindex.py
@@ -137,10 +136,3 @@ Gradually move work items from here to repo Issues.
 
 ### P2 - Medium Priority
 - **[P2, medium]** Use pydantic.ai for model drivers, to support non-openai models
-
----
-
-## Architecture Decisions
-
-### P2 - Medium Priority
-- **[P2, medium]** Keep in-memory version (with some compromises) for comparison
--- a/test/test_conversation_metadata.py
+++ b/test/test_conversation_metadata.py
@@ -3,26 +3,34 @@
 
 """Tests for conversation metadata operations in SQLite storage provider."""
 
-import os
-import tempfile
-from collections.abc import AsyncGenerator, Generator
+import asyncio
+from collections.abc import AsyncGenerator
 from dataclasses import field
 from datetime import datetime, timezone
+import os
+import sqlite3
+import tempfile
+import time
 
 import pytest
 import pytest_asyncio
 from pydantic.dataclasses import dataclass
 
-from typeagent.aitools.embeddings import AsyncEmbeddingModel
+from typeagent.aitools.embeddings import AsyncEmbeddingModel, TEST_MODEL_NAME
 from typeagent.aitools.vectorbase import TextEmbeddingIndexSettings
 from typeagent.knowpro.convsettings import (
+    ConversationSettings,
     MessageTextIndexSettings,
     RelatedTermIndexSettings,
 )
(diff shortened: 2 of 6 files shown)


typeagent/knowpro/knowledge.py
```diff
@@ -36,14 +37,16 @@ async def extract_knowledge_from_text_batch(
...
-    return results
+    semaphore = asyncio.Semaphore(concurrency)
+
+    async def extract_with_semaphore(text: str) -> Result[kplib.KnowledgeResponse]:
+        async with semaphore:
+            return await extract_knowledge_from_text(
+                knowledge_extractor, text, max_retries
+            )
+
+    tasks = [asyncio.create_task(extract_with_semaphore(text)) for text in text_batch]
```
gvanrossum: @KRRT7 -- This may create an excessive number of tasks when a large batch is presented (it's totally legit to have a single batch of all messages). I'd propose using a queue with a size limit instead, in addition to the semaphore.

async with asyncio.taskgroup():
  q = asyncio.Queue(max=2*concurrency + 2)
  for text in text_batch:
    await q.push(text)

and a corresponding pool of workers that read from the queue.

KRRT7:
@gvanrossum I don't know if I've ever implemented a prod - consumer pattern, can you take a look?

typeagent/knowpro/knowledge.py
```diff
@@ -29,20 +30,54 @@ async def extract_knowledge_from_text(
...
+async def batch_worker(
+    q: asyncio.Queue[tuple[int, str] | None],
+    knowledge_extractor: IKnowledgeExtractor,
+    results: list[Result[kplib.KnowledgeResponse] | None],
+    max_retries: int,
+) -> None:
+    while True:
+        item = await q.get()
+        if item is None:
+            return
```
gvanrossum:
```suggestion
    while item := await q.get():
```

typeagent/knowpro/knowledge.py
```diff
@@ -29,20 +30,54 @@ async def extract_knowledge_from_text(
...
+    if not text_batch:
+        return []
+
+    q: asyncio.Queue[tuple[int, str] | None] = asyncio.Queue(
+        maxsize=2 * concurrency + 2
+    )
+    results: list[Result[kplib.KnowledgeResponse] | None] = [None] * len(text_batch)
+
+    async with asyncio.TaskGroup() as tg:
+        tg.create_task(batch_producer(q, text_batch, concurrency))
```
gvanrossum: You could create the workers *first* (they'll all just be waiting for something to enter the queue) and have the loop that stuffs the queue written out here, without making it a separate task.

I'm also okay with making the batch worker a local function again; that should save some arguments (though not necessarily nanoseconds :-). Up to you.

gvanrossum:
Also you have introduced a type error:

```txt
  /home/runner/work/typeagent-py/typeagent-py/typeagent/knowpro/knowledge.py:81:12 - error: Type "list[Result[KnowledgeResponse] | None]" is not assignable to return type "list[Result[KnowledgeResponse]]"
    "list[Result[KnowledgeResponse] | None]" is not assignable to "list[Result[KnowledgeResponse]]"
      Type parameter "_T@list" is invariant, but "Result[KnowledgeResponse] | None" is not the same as "Result[KnowledgeResponse]"
      Consider switching from "list" to "Sequence" which is covariant (reportReturnType)
```

KRRT7:
I'm not entirely it's a true type errors ince all items are populated by workers before returning (due to TaskGroup waiting for all tasks), the None values are temporary scaffolding, but type checker doesn't know this. I could use cast for this but some people don't like that.

gvanrossum:
> I'm not entirely it's a true type errors ince all items are populated by workers before returning (due to TaskGroup waiting for all tasks), the None values are temporary scaffolding, but type checker doesn't know this. I could use cast for this but some people don't like that.

We must pass the static type checker. I think it's at least worth adding an assert like this:
```py
assert not any(result is None for result in results)
```

That won't fix the static type error (alas) but with that assert I'm fine with either adding `# type: ignore` or a cast.

KRRT7:
I ended up solving it another way,  no more type checker complains

gvanrossum:
(Sorry, somehow didn't send these comments earlier.)

typeagent/knowpro/knowledge.py
KRRT7: opted for the first option

PR #90, opened 2025-11-13
Move podcast ingestion main() to new file tools/ingest_podcast.py

gvanrossum:
Also catch RuntimeError and ValueError and print just the error, suppressing the traceback. This seems more user-friendly (?). To disable, edit the source code, e.g. to put plain `raise` before the `print()` call.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -14,19 +14,21 @@ Gradually move work items from here to repo Issues.
 
 ---
 
-## Storage & Persistence
-
-### P0 - Critical
-- **[P0, large]** Fix all bugs related to ordinals/ids relying on starting at 0 and no gaps (prepare for deletions)
+## Additional code reviews (of AI-vibe-coded things)
 
 ### P1 - High Priority
 - **[P1, medium]** Scrutinize sqlite/reltermsindex.py
 - **[P1, medium]** Review the new storage code more carefully, adding notes
+
+## Storage & Persistence
+
+### P1 - High Priority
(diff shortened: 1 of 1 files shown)


tools/ingest_podcast.py
```diff
@@ -0,0 +1,85 @@
+import argparse
+import asyncio
+import os
+
+from typeagent.aitools.utils import load_dotenv
+from typeagent.knowpro.convsettings import ConversationSettings
+from typeagent.podcasts.podcast_ingest import ingest_podcast
+
+DEFAULT_TRANSCRIPT = "testdata/Episode_53_AdrianTchaikovsky.txt"
+CHARS_PER_MINUTE = 1050  # My guess for average speech rate incl. overhead
```
robgruen, line 10: could make CHARS_PER_MINUTE a parameter if you're feeling generous

gvanrossum: I'm just trying to get approximately the same timings as the TS ingest code, which asks for the length of the recording in minutes. Either way it's pretty bogus because it assumes a constant rate of time per character. The VTT files have accurate timing -- but they don't have "who's in the scene" information. And in order to convert Adrian T. to VTT we'd have to make up timings anyway. We *could* do that and do the "listener" information the way we do it for podcasts, using a flag, but I'm feeling there's not enough value to doing better on any of this, so I'm just keeping things alive with minimal effort...

Issue #91, opened 2025-11-14
some tests fail on cmd line, but it works in VS Code Test Explorer


bmerkle:
running tests in windows powershell fails 6 tests unfortunately (example) below.
running tests with VS Code Test Explorer runs all tests successfully.

seems to be a environment problem, (googling or asking chatGPT leads to same direction) but i did not find a real solution or trace.
python interpreter and env variables are the same

not a big issue but i usually prefer cmd line over clicking ;) and github ci obviously runs fine on linux ubuntu

```
(.venv) PS C:\work\microsoft\typeagent-py\test> .\make.bat test
=================================================== short test summary info ==================================================== 
FAILED test_factory.py::test_create_conversation_and_add_messages - RuntimeError: Knowledge extraction failed: Expecting value: line 1 column 1 (char 0) -- MESSAGE='Hello, how are you?'
FAILED test_incremental_index.py::test_incremental_index_with_vtt_files - RuntimeError: Failed to parse VTT file testdata/Confuse-A-Cat.vtt: [Errno 2] No such file or directory: 'testdata/Confuse-A-...
FAILED test_mcp_server.py::test_mcp_server_query_conversation_slow - JSONDecodeError('Expecting value: line 1 column 1 (char 0)') [single exception in ExceptionGroup]
FAILED test_podcasts.py::test_ingest_podcast - FileNotFoundError: [Errno 2] No such file or directory: 'testdata/FakePodcast.txt'
FAILED test_query_method.py::test_query_method_basic - RuntimeError: Knowledge extraction failed: Expecting value: line 1 column 1 (char 0) -- MESSAGE='Welcome to the Python progr...
FAILED test_query_method.py::test_query_method_empty_conversation - AssertionError: assert ('no answer' in 'search failed: expecting value: line 1 column 1 (char 0)' or 'unable to find' in 'se...
========================================== 6 failed, 345 passed, 4 skipped in 48.54s =========================================== 
```

gvanrossum:
Hm, all of these pass for me with the latest main. Maybe try `make clean` and `make venv`? (where "make" is .\make.bat)

gvanrossum:
@bmerkle - if that doesn't fix it, can you post the full output from running at least one of the failing tests? Also, what Windows version?

bmerkle:
below is
- short output of envrionment
- longer output of test trace
- screenshot of Test Exporer (only 1 test fails)
- git is on latest main. 

[...]

PR #92, opened 2025-11-14
Fix cleanup in temp_db_path fixture on Windows

gvanrossum:
The failure is one of those annoying Windows things where the virus scanner isn't done scanning a file you just wrote when you're trying to delete it. Since it's a temp file, I think leaking the file is acceptable.

See #91.

### DIFF
--- a/test/fixtures.py
+++ b/test/fixtures.py
@@ -75,9 +75,14 @@ def temp_db_path() -> Iterator[str]:
     temp_file = tempfile.NamedTemporaryFile(suffix=".sqlite", delete=False)
     path = temp_file.name
     temp_file.close()
+
     yield path
+
     if os.path.exists(path):
-        os.remove(path)
+        try:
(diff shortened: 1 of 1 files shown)


KRRT7:
why not just chuck it to tempfile.TemporaryDirectory() and let the OS handle it? when I had this same problem this solved my issues

gvanrossum-ms:
> why not just chuck it to tempfile.TemporaryDirectory() and let the OS handle it? when I had this same problem this solved my issues

Hm, I just tried that but it doesn't work. Same error message as on main.

So I'm going to merge this.

KRRT7:
https://github.com/codeflash-ai/codeflash/blob/c821e6763e701a4309fdb41d6ca82c1fce84558b/tests/test_cmd_init.py#L15-L18

not sure why it didn't work for you, it works fine for me elsewhere

PR #103, opened 2025-11-18
More fixes related to get_multiple

gvanrossum-ms:
The "fix" for `get_multiple()` in #102 was incorrect (it returned items sorted by `semref_id` instead of corresponding to `arg`). Moreover, the tests I added were testing the messages collection, not the `semrefs` collection.

This PR fixes semref `get_multiple()` to return items in the right order (@robgruen: same solution as the .NET version!). It also optimizes messages `get_multiple()` to use the same approach. It also adds more comprehensive tests for `get_multiple()` and some other `get_` methods and `__aiter__`, *and* comprehensive tests for semref methods. The new tests also require `get_multiple()` to raise `IndexError` when any ordinal in `arg` is out of range -- this required adding code to enforce that in the memory collection code.

### DIFF
--- a/test/test_storage_providers_unified.py
+++ b/test/test_storage_providers_unified.py
@@ -95,7 +95,7 @@ async def test_all_index_creation(
     storage_provider_type: tuple[IStorageProvider, str], needs_auth: None
 ):
     """Test that all 6 index types are created and accessible in both providers."""
-    storage_provider, provider_type = storage_provider_type
+    storage_provider, _ = storage_provider_type
 
     # Test all index types are created and return proper interface objects
     conv_index = await storage_provider.get_semantic_ref_index()
@@ -128,7 +128,7 @@ async def test_index_persistence(
     storage_provider_type: tuple[IStorageProvider, str], needs_auth: None
 ):
     """Test that same index instance is returned across calls in both providers."""
-    storage_provider, provider_type = storage_provider_type
+    storage_provider, _ = storage_provider_type
 
     # All index types should return same instance across calls
     conv1 = await storage_provider.get_semantic_ref_index()
@@ -149,7 +149,7 @@ async def test_message_collection_basic_operations(
     storage_provider_type: tuple[IStorageProvider, str], needs_auth: None
 ):
     """Test basic message collection operations work identically in both providers."""
-    storage_provider, provider_type = storage_provider_type
+    storage_provider, _ = storage_provider_type
 
     # Create message collection
     collection = await storage_provider.get_message_collection()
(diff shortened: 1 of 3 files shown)


robgruen (approved):

gvanrossum-ms:
NOTE: There's an open question. Should get_multiple raise an exception if any ordinal is out of range? Or should it return a list containing None for ordinals that aren't present in the collection? Let's discuss this before reviewing this PR. Note that get_item(i) raises IndexError when i is out of range (or absent for some other reason, like when we implement deletions).

The TS code returns undefined from getItem (sp?) and an array containing undefined for absent indexes. This is consistent. I don't think I want get_item() to return None for absent indices; but I'm not sure that it follows that get_multiple() should raise for absent indices -- it's more useful to return a list containing None for absent indices...

CONCLUSION: I talked this through with @robgruen and he convinced me that raising `IndexError` is the right thing -- *for now*. In the future we may want to return a list with `None` indicating missing items, but we may want a flag to control the behavior. Missing items may be permanently missing (e.g., permanently deleted) or temporarily gone only -- e.g., when a database shard is down or we're in the middle of GC or "dreaming".

test/test_storage_providers_unified.py
```diff
@@ -680,35 +696,62 @@ async def test_collection_operations_comprehensive(
...
     no_result = await message_collection.get_multiple([])
     assert len(no_result) == 0
-
-    # Test iteration
-    collection_list = [item async for item in message_collection]
-    assert len(collection_list) == 3
-    for i, msg in enumerate(collection_list):
-        assert msg.text_chunks == test_messages[i].text_chunks
+    # Edge case: duplicate index
+    result = await message_collection.get_multiple([0, 0, 1])
```
robgruen, line 747: in KnowPro.NET we eliminate the duplicates.  Should we standardize?

gvanrossum: But then the output doesn't correspond to the index. For messages, at least, the returned message doesn't include its msg_id (it doesn't exist in the type definition in interfaces.py). I believe that we should *always* be able to do
```py
args = [....]
msgs = msg_coll.get_multiple(args)
d = dict(zip(args, msgs))
for arg in args:
    print(arg, ":", d[arg])
```
If you eliminate duplicates, this will break subtly: the code runs, but will print incorrect results. E.g. for args = [1, 2, 2, 3] it will print
```txt
1: <message 1>
2: <message 3>
```
(Note that zip stops when any of its input iterators end, so zip([1, 2, 2, 3], [msg1, msg2, msg3]) returns [(1, msg1), (2, msg2), (2, msg3)]. And dict() keeps the last value for any given key.

PR #104, opened 2025-11-18
Add install-libatomic target to Makefile

### LINKED ISSUE #96: make venv fails on clean Ubunty distro with missing shared libraries: libatomic.so.1
Small issue when I **setup on a complete new Ubunt installation**:

```
merkle@merkle-laptop:~/typeagent-py$ make venv
(If 'uv' fails with 'No such file or directory', try 'make install-uv')
uv sync -q --extra dev
black 25.9.0 (compiled: yes)
Python (CPython) 3.12.3
WARNING: there is a new pyright version available (v1.1.406 -> v1.1.407).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

 * Install prebuilt node (25.2.0) ..... done.
/home/merkle/.cache/pyright-python/nodeenv/bin/node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
make: *** [Makefile:63: .venv] Error 127
```

steps to reproduce:
- install ubuntu via wsl.exe
- install uv (curl -LsSf https://astral.sh/uv/install.sh | sh)
- install make (sudo apt install make)
- try to setup typeagent-py (make venv)

error: message above the libatomic.so is missing.

not a big issue, but we might to have assert that libatomic.so.1 is present a-prior.

the reason is: 
Node.js, which is used from Pyright can not find libatomic.so.1 . 
libatomic (provided usually by GCC ) is often missin on minimal linux systems, (docker) containers or WSL.

possible workaround: 
- check if libatomic is there (ldconfig -p | grep libatomic)
- if not, install it (sudo apt install libatomic1)

i can try to submit a PR :-)

bmerkle:
Fixes #96.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,21 +64,14 @@ venv: .venv
 	@.venv/bin/pyright --version
 	@.venv/bin/pytest --version
 
+.PHONY: install-uv
 install-uv:
 	curl -Ls https://astral.sh/uv/install.sh | sh
 
+.PHONY: install-libatomic
 install-libatomic:
-	@if ldconfig -p | grep -q libatomic.so.1; then \
-		echo "libatomic already installed"; \
-	else \
-		echo "Installing libatomic..."; \
-		if command -v apt-get >/dev/null 2>&1; then \
-			sudo apt-get update && sudo apt-get install -y libatomic1; \
-		else \
-			echo "No supported package manager found. Please install libatomic manually."; \
-			exit 1; \
-		fi \
-	fi
+	sudo apt-get update
+	sudo apt-get install -y libatomic1
 
 .PHONY: clean
 clean:


Makefile
```diff
@@ -60,12 +60,26 @@ venv: .venv
 	@echo "(If 'uv' fails with 'No such file or directory', try 'make install-uv')"
 	uv sync -q --extra dev
 	@.venv/bin/black --version | sed 's/, / /'
+	@echo "(If 'pyright' fails with 'error while loading shared libraries: libatomic.so.1:', try 'make install-libatomic')"
```
gvanrossum, line 63: This sounds like the right level of support.

Makefile
```diff
@@ -60,12 +60,26 @@ venv: .venv
...
+	else \
+		echo "Installing libatomic..."; \
+		if command -v apt-get >/dev/null 2>&1; then \
+			sudo apt-get update && sudo apt-get install -y libatomic1; \
+		else \
+			echo "No supported package manager found. Please install libatomic manually."; \
+			exit 1; \
+		fi \
+	fi
+
```
gvanrossum, line 75: Maybe we can replace this with a much simpler command?
```
install-libatomic:
    sudo apt-get update
    sudo apt-get install -y libatomic1
```

If these fails it's up to the user to figure out what else to do.

If we run this while libatomic1 has already been installed, it's a quick no-op. Nothing depends on this target.

I do think that both install-uv and install-libatomic should have a .PHONY: <target> prefix.

gvanrossum (changes requested):

Makefile
bmerkle: thanks for feedback. makes sense. should be done

gvanrossum (approved):

PR #106, opened 2025-11-21
Setup build pipeline to run online tests

robgruen:
- Updated github build action to log into azure and dynamically get secrets
- Set permissions on build action appropriately so it only runs when triggered by authorized user
- Added getKeys.py to get keys from keyvault (ported from [getKeys](https://github.com/microsoft/TypeAgent/blob/main/ts/tools/scripts/getKeys.mjs) using Claude Sonnet 4.5).

### DIFF
--- a/tools/getKeys.py
+++ b/tools/getKeys.py
@@ -475,8 +475,8 @@ async def push_secret(
 
     if secret_key in secrets:
         print(f"  {secret_key} changed.")
-        print(f"    Current value: {secret_value}")
-        print(f"    New value: {value}")
+        print(f"    Current value: [REDACTED]")
+        print(f"    New value: [REDACTED]")
         answer = input(
             f"  Are you sure you want to overwrite the value of {secret_key}? (y/n): "
         )


pyproject.toml
```diff
@@ -30,6 +30,8 @@ classifiers = [
 ]
 dependencies = [
   "azure-identity>=1.22.0",
+  "azure-mgmt-authorization>=4.0.0",
+  "azure-mgmt-keyvault>=12.1.1",
```
gvanrossum, line 34: Maybe these two belong in the extras section? They aren't needed for people who just install the package.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
+#!/usr/bin/env python3
```
gvanrossum, line 1: Filename should use Python convention, get_keys.py, and ditto for config file.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+import os
+import sys
+import json
+import subprocess
+import argparse
+import asyncio
+import uuid
+from pathlib import Path
+from typing import Dict, List, Tuple, Optional, Set
+from datetime import datetime
```
gvanrossum, line 14: Alphabetize imports.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+
+# Azure SDK imports
+from azure.identity import DefaultAzureCredential
+from azure.mgmt.authorization import AuthorizationManagementClient
+
+
+# ANSI color codes for terminal output
```
gvanrossum, line 21: Use colorama instead -- already used elsewhere.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+            )
+
+            print(result)
+            print(
+                colored(
+                    f"ELEVATION SUCCESSFUL for role {options['roleName']}", Colors.GREEN
+                )
+            )
+        except Exception as e:
+            print(e)
```
gvanrossum, line 162: I'd say `print(repr(e))` so the exception class names shown too.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+                )
+                return role_id
+
+        print(
+            colored(
+                f"ERROR: Unable to find the requested role '{role_name}'. Are you certain you are logged into the correct subscription?",
+                Colors.RED,
+            )
+        )
+        raise Exception(f"Unable to find the role '{role_name}'.")
```
gvanrossum, line 209: Use a more specific exception name, e.g. RuntimeError.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+            result = subprocess.run(
+                [AZ_COMMAND, "ad", "signed-in-user", "show"],
+                capture_output=True,
+                text=True,
+                check=True,
+            )
+            account_details = json.loads(result.stdout)
+            return account_details["id"]
+        except Exception as e:
+            print(e)
```
gvanrossum, line 223: Again

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+            if not continue_on_failure:
+                print(
+                    colored(
+                        "ERROR: Unable to get principal id of the current user.",
+                        Colors.RED,
+                    )
+                )
+                sys.exit(12)
```
gvanrossum, line 231: What does 12 mean? Normally we use 1.

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+                    "continueOnFailure": True,
+                }
+            )
+
+            print(colored("Elevation successful.", Colors.GREEN))
+            print(colored("Waiting 5 seconds...", Colors.YELLOW))
+            await asyncio.sleep(5)
+
+            return client.get_secrets(vault_name)
+        except Exception:
```
gvanrossum, line 352: Not printing the exception at all?

tools/getKeys.py
```diff
@@ -0,0 +1,724 @@
...
+                    "expirationType": "AfterDuration",
+                    "expirationDuration": "PT5M",  # 5 minutes
+                    "continueOnFailure": True,
+                }
+            )
+
+            print(colored("Elevation successful.", Colors.GREEN))
+            print(colored("Waiting 5 seconds...", Colors.YELLOW))
+            await asyncio.sleep(5)
+        except Exception:
```
gvanrossum, line 377: Again

PR #108, opened 2025-11-23
Add --force flag to tools/release.py and remove 'except Exception'

gvanrossum:
Sometimes you want to push an earlier release or while the repo is dirty. Use `--force` to disable the errors for those cases.

Remove two `try:` // `except Exception as e: print(e); return` blocks, they don't help anything (expected exceptions aren't handled this way, and for unexpected ones I strongly prefer to get the full traceback.

### DIFF
--- a/tools/release.py
+++ b/tools/release.py
@@ -6,13 +6,13 @@
 Release automation script for the TypeAgent Python package.
 
 This script:
-1. Bumps the patch version (3rd part) in pyproject.toml, or sets it to a specified version
+1. Bumps the patch version (3rd part) in pyproject.toml, or sets the whole version
 2. Commits the change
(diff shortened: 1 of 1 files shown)


tools/release.py
```diff
@@ -246,97 +241,100 @@ def main():
...
+        current_tuple = (current_major, current_minor, current_patch)
+        new_tuple = (new_major, new_minor, new_patch)
+        comparison = compare_versions(new_tuple, current_tuple)
 
-            if comparison <= 0:
+        if comparison <= 0:
+            if args.force:
                 print(
-                    f"Error: New version {args.version} must be higher than current version {current_version}",
+                    f"Warning: New version {args.version} precedes current version {current_version}. (forced)"
```
robgruen: nit: "matches or precedes"

PR #111, opened 2025-12-03
New tool: ingest_email.py

gvanrossum-ms:
This does the same as `@add_messages` in test_gmail.py does, but now the storage provider interface has changed to allow storing the "is it ingested" flag per message id in the database, so it is transactionally safe.

(Almost all by Claude Opus 4.5 Preview.)

I'm not in  a hurry to remove test_email.py -- we need to make sure that all its use cases are now incorporated into either ingest_email.py ro query.py. I think @parse_messages is the only one left.

### DIFF
--- a/tools/ingest_email.py
+++ b/tools/ingest_email.py
@@ -70,14 +70,16 @@ def collect_email_files(paths: list[str], verbose: bool) -> list[Path]:
             if path.suffix.lower() == ".eml":
                 email_files.append(path)
             else:
-                print(f"Warning: Skipping non-.eml file: {path}", file=sys.stderr)
+                print(f"Error: Skipping non-.eml file: {path}", file=sys.stderr)
+                sys.exit(1)
         elif path.is_dir():
-            eml_files = list(path.glob("*.eml"))
+            eml_files = sorted(path.glob("*.eml"))
             if verbose:
                 print(f"  Found {len(eml_files)} .eml files in {path}")
             email_files.extend(eml_files)
         else:
-            print(f"Warning: Skipping special file: {path}", file=sys.stderr)
+            print(f"Error: Not a file or directory: {path}", file=sys.stderr)
+            sys.exit(1)
 
     return email_files
 


tools/ingest_email.py
```diff
@@ -0,0 +1,241 @@
...
+        path = Path(path_str)
+        if not path.exists():
+            print(f"Error: Path '{path}' not found", file=sys.stderr)
+            sys.exit(1)
+
+        if path.is_file():
+            if path.suffix.lower() == ".eml":
+                email_files.append(path)
+            else:
+                print(f"Warning: Skipping non-.eml file: {path}", file=sys.stderr)
```
gvanrossum: Should also error.

tools/ingest_email.py
```diff
@@ -0,0 +1,241 @@
...
+        elif path.is_dir():
+            eml_files = list(path.glob("*.eml"))
```
gvanrossum: Sorted

tools/ingest_email.py
```diff
@@ -0,0 +1,241 @@
...
+            # Print progress periodically
+            if not verbose and (i + 1) % batch_size == 0:
+                elapsed = time.time() - start_time
+                semref_count = await semref_coll.size()
+                print(
+                    f"  [{i + 1}/{len(email_files)}] {successful_count} imported | "
+                    f"{semref_count} refs | {elapsed:.1f}s elapsed"
+                )
+
+        except Exception as e:
```
gvanrossum, line 189: Drop the exception handling

tools/ingest_email.py
robgruen: Also for convenience, do we want to consider adding a recursive argument and then use rglob()?

tools/ingest_email.py
```diff
@@ -0,0 +1,241 @@
...
+
+async def ingest_emails(
+    paths: list[str],
+    database: str,
+    verbose: bool = False,
+) -> None:
+    """Ingest email files into a database."""
+
+    # Collect all .eml files
+    if verbose:
```
robgruen, line 95: Is there a cleaner way to do this verbose printing?  Could there be a printVerbose() method that we call instead of print() and then it decides to print the message or not based on the verbose flag?

tools/ingest_email.py
```diff
@@ -0,0 +1,241 @@
...
+                print(f"    Body chunks: {len(email.text_chunks)}")
+                for chunk in email.text_chunks:
+                    # Show first 200 chars of each chunk
+                    preview = chunk[:200].replace("\n", " ")
+                    if len(chunk) > 200:
+                        preview += "..."
+                    print(f"      {preview}")
+
+            # Pass source_id to mark as ingested atomically with the message
+            source_ids = [email_id] if email_id else None
```
robgruen, line 174: maybe emit a warning when the email id doesn't exist.  

Do we try to create one based on the from/to/timestamp/subject hash if we don't have one otherwise?

typeagent/storage/sqlite/provider.py
```diff
@@ -619,3 +619,36 @@ def update_conversation_timestamps(
...
+        cursor = self.db.cursor()
+        cursor.execute(
+            "SELECT 1 FROM IngestedSources WHERE source_id = ?", (source_id,)
+        )
+        return cursor.fetchone() is not None
+
+    def mark_source_ingested(self, source_id: str) -> None:
+        """Mark a source as ingested.
+
+        This performs an INSERT but does NOT commit. It should be called within
```
robgruen, line 643: I haven't tried this, but the docs say you can do cursor.in_transaction to see if there's an active transaction.  That could keep someone from calling this if there's no active transaction.

tools/ingest_email.py
gvanrossum-ms: Actually, no. It serves a purpose (report as not indexed and continue with next message).

tools/ingest_email.py
gvanrossum-ms: Let's not -- until there is actual demand, it will be easy to add then.

tools/ingest_email.py
gvanrossum: It would feel a little "clever", so I'll skip it.

tools/ingest_email.py
gvanrossum: We could. I'll defer doing that until there's demand.

typeagent/storage/sqlite/provider.py
gvanrossum: Meh, this function only exists so add_messages_with_indexing can call it. I'm not too worried about users calling it wrongly.

Issue #112, opened 2025-12-03
Code Quality & Cleanup (from TODO.md)


bmerkle:
In this issue we trac possible Code Quality & Cleanup Items from TODO.md

### P1 - High Priority
- **[P1, large]** Make coding style more uniform (e.g. docstrings)
- **[P1, large]** Reduce code size

### P2 - Medium Priority
- **[P2, small]** Avoid most inline imports
- **[P2, medium]** Break cycles by moving things to their own file if necessary
- **[P2, medium]** Unify or align or refactor `VectorBase` and `EmbeddingIndex`
- **[P2, medium]** Address `TODO` comments (too numerous)
- **[P2, medium]** Address `raise NotImplementedError("TODO")` (five found) -- implement it

### P3 - Low Priority
- **[P3, medium]** Change inconsistent module names (Claude uses different naming style)
- **[P3, medium]** Rewrite podcast parsing without regexes
- **[P3, medium]** Switch from Protocol to ABC
- **[P3, medium]** Reduce duplication between ingest_vtt.py and typeagent/transcripts/
- **[P3, small]** Rename `kplib.py` to something ending in `_schema.py`

bmerkle:
> If you want to try your hand at some items from TODO.md, maybe these make decent starter issues:
> 
> ## Code Quality & Cleanup
> ### P2 - Medium Priority
> - **[P2, small]** Avoid most inline imports
> - **[P2, medium]** Break cycles by moving things to their own file if necessary
> 
> I'd start with the former but back out whenever moving an inline import reveals a circular dependency -- for those we often need to do the latter, which is much more subtle and requires a deep understanding of the architecture in order to move the *right* things into separate files.
> 
> I'm not in a hurry with the development infra items (all P3s anyways). Feel free to document the test/build process. The release process I'd rather document myself because I think there are actually some improvements needed for that. 

 _Originally posted by @gvanrossum in [#45](https://github.com/microsoft/typeagent-py/issues/45#issuecomment-3555273164)_

bmerkle:
> For moving inline imports to the top of the file, I'd try one file at a time under typechat. There are some traps where these are needed to avoid circular imports, but many are undoubtedly due to Claude's preference for that style, which is not mine.
> 
> For the tests I think you should be able to move all imports out, since there are no circular imports between tests and everything else ought to be importable in any order.
> 
> ...
> 
[...]

PR #113, opened 2025-12-03
Refactoring inline imports and resolve circular dependencies

bmerkle:
Refactoring for issue https://github.com/microsoft/typeagent-py/issues/112

all inline imports are removed

formatted imports and from imports
(sorted, arranged etc, via ruff tool)  ruff check --select I --fix .

circular dependecies are removed, a new types.py file was introduced

searchlib.py improved handling of pydantic classes

note that the patch set is quite large (99 files) but i looked through to assert that nothing broke.

unit tests ran fine. (except mcp server)

### DIFF
--- a/typeagent/storage/memory/provider.py
+++ b/typeagent/storage/memory/provider.py
@@ -39,6 +39,7 @@ class MemoryStorageProvider[TMessage: IMessage](IStorageProvider[TMessage]):
     _message_text_index: MessageTextIndex
     _related_terms_index: RelatedTermsIndex
     _conversation_threads: ConversationThreads
+    _ingested_sources: set[str]
 
     def __init__(
         self,
@@ -58,6 +59,7 @@ class MemoryStorageProvider[TMessage: IMessage](IStorageProvider[TMessage]):
         self._related_terms_index = RelatedTermsIndex(related_terms_settings)
         thread_settings = message_text_settings.embedding_index_settings
         self._conversation_threads = ConversationThreads(thread_settings)
+        self._ingested_sources = set()
 
     async def __aenter__(self) -> "MemoryStorageProvider[TMessage]":
         """Enter transaction context. No-op for in-memory storage."""
@@ -132,3 +134,23 @@ class MemoryStorageProvider[TMessage: IMessage](IStorageProvider[TMessage]):
             created_at: Optional creation timestamp (ignored)
             updated_at: Optional last updated timestamp (ignored)
         """
+        pass
+
+    def is_source_ingested(self, source_id: str) -> bool:
+        """Check if a source has already been ingested.
+
+        Args:
+            source_id: External source identifier (email ID, file path, etc.)
+
+        Returns:
+            True if the source has been ingested, False otherwise.
+        """
+        return source_id in self._ingested_sources
+
+    def mark_source_ingested(self, source_id: str) -> None:
+        """Mark a source as ingested.
+
+        Args:
+            source_id: External source identifier (email ID, file path, etc.)
+        """
+        self._ingested_sources.add(source_id)
--- a/typeagent/storage/sqlite/provider.py
+++ b/typeagent/storage/sqlite/provider.py
@@ -608,3 +608,36 @@ class SqliteStorageProvider[TMessage: interfaces.IMessage](
     def get_db_version(self) -> int:
         """Get the database schema version."""
         return get_db_schema_version(self.db)
+
+    def is_source_ingested(self, source_id: str) -> bool:
+        """Check if a source has already been ingested.
+
+        This is a read-only operation that can be called outside of a transaction.
+
+        Args:
+            source_id: External source identifier (email ID, file path, etc.)
+
+        Returns:
+            True if the source has been ingested, False otherwise.
+        """
+        cursor = self.db.cursor()
+        cursor.execute(
+            "SELECT 1 FROM IngestedSources WHERE source_id = ?", (source_id,)
+        )
+        return cursor.fetchone() is not None
+
+    def mark_source_ingested(self, source_id: str) -> None:
+        """Mark a source as ingested.
+
+        This performs an INSERT but does NOT commit. It should be called within
+        a transaction context (e.g., inside `async with storage_provider:`).
+        The commit happens when the transaction context exits successfully.
+
+        Args:
+            source_id: External source identifier (email ID, file path, etc.)
+        """
+        cursor = self.db.cursor()
+        cursor.execute(
+            "INSERT OR IGNORE INTO IngestedSources (source_id) VALUES (?)",
+            (source_id,),
+        )
--- a/docs/demos.md
+++ b/docs/demos.md
@@ -30,18 +30,20 @@ python tools/query.py -d mp.db
 ```
 (You just type questions and it prints answers.)
 
-## How we did the GMail demo
+## How we did the Gmail demo
 
 The demo consisted of loading a large number (around 500) email messages
 into a database, and querying the database about those messages.
 The loading (ingestion) process was done ahead as it takes a long time.
 
-We used the GMail API to download 550 messages from Guido's GMail
+We used the Gmail API to download 550 messages from Guido's Gmail
 (details below).
 
 Given a folder with `*.eml` files in MIME format, we ran our email
 ingestion tool, `tools/test_email.py`. (All these details will change
 in the future, hopefully to be more similar to `ingest_vtt.py`.)
+
+**TODO: Switch to describing ingest_email.py.**
 
 The tool takes one positional argument, a directory, in which it will
 create a SQLite database named `gmail.db`.
@@ -60,16 +62,32 @@ next file.
 We can then query the `gmail.db` database using the same `query.py`
 tool that we used for the Monty Python demo.
 
-### How to use the GMail API to download messages
+### How to use the Gmail API to download messages
 
 In the `gmail/` folder you'll find a tool named `gmail_dump.py` which
-will download any number of messages (default 50) using the GMail API.
-In order to use the GMail API, however, you have to create a
+will download any number of messages (default 50) using the Gmail API.
+In order to use the Gmail API, however, you have to create a (free)
 Google Cloud app and configure it appropriately.
 
-In order to figure out how to set up the (free) Google Cloud app we
-used the instructions at [GeeksForGeeks
-](https://www.geeksforgeeks.org/devops/how-to-create-a-gcp-project/).
+We created  created an app in test mode at
+[Google Cloud Console](https://console.cloud.google.com) and gave it
+access to the Gmail API (I forget how exactly we did this part).
+
+To create the needed client secret, we navigated to Client (side bar)
+and clicked on "+ Create Client" (in the row of actions at the top),
+selected "Desktop app", gave it a name, hit Create, scrolled down in the
+resulting dialog box, and hit "Download JSON". This produced a JSON file
+which should be copied into _client_secret.json_ in the gmail folder.
+(The Cloud Console interface may look different for you.)
+
+The first time you run the gmail_dump.py script, it will take you to
+a browser where you have to log in and agree to various warnings about
+using an app in test mode etc. The gmail_dump.py script then writes a
+file _token.json_ and you're good for a week or so. When token.json
+expires, unfortunately you get a crash and you have to manually delete
+it to trigger the login flow again.
+(Sometimes starting a browser may fail, e.g. under WSL. Take the URL
+that's printed and manually go there.)
 
 The rest of the email ingestion pipeline doesn't care where you got
 your `*.eml` files from -- every email provider has its own quirks.
--- a/docs/gmail.md
+++ b/docs/gmail.md
@@ -6,3 +6,7 @@ Until we have time to write this up, your best bet is to
 ask your favorite search engine or LLM-based chat bot for help.
 
 More TBD.
+In the meantime there are some hints in [Demos][def].
+
+
+[def]: demos.md#how-to-use-the-gmail-api-to-download-messages
--- a/tools/ingest_email.py
+++ b/tools/ingest_email.py
@@ -1,0 +1,243 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""
+Email Ingestion Tool
+
+This script ingests email (.eml) files into a SQLite database
+that can be queried using tools/query.py.
+
+Usage:
+    python tools/ingest_email.py -d email.db inbox_dump/
+    python tools/ingest_email.py -d email.db message1.eml message2.eml
+    python query.py --database email.db --query "What was discussed?"
+"""
+
+import argparse
+import asyncio
+import os
+import sys
+import time
+from pathlib import Path
+
+from typeagent.aitools import utils
+from typeagent.emails.email_import import import_email_from_file
+from typeagent.emails.email_memory import EmailMemory
+from typeagent.emails.email_message import EmailMessage
+from typeagent.knowpro.convsettings import ConversationSettings
+from typeagent.storage.utils import create_storage_provider
+
+
+def create_arg_parser() -> argparse.ArgumentParser:
+    """Create argument parser for the email ingestion tool."""
+    parser = argparse.ArgumentParser(
+        description="Ingest email (.eml) files into a database for querying",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+
+    parser.add_argument(
+        "paths",
+        nargs="+",
+        help="Path to one or more .eml files or directories containing .eml files",
+    )
+
+    parser.add_argument(
+        "-d",
+        "--database",
+        required=True,
+        help="Path to the SQLite database file to create/use",
+    )
+
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", help="Show verbose/debug output"
+    )
+
+    return parser
+
+
+def collect_email_files(paths: list[str], verbose: bool) -> list[Path]:
+    """Collect all .eml files from the given paths (files or directories)."""
+    email_files: list[Path] = []
+
+    for path_str in paths:
+        path = Path(path_str)
+        if not path.exists():
+            print(f"Error: Path '{path}' not found", file=sys.stderr)
+            sys.exit(1)
+
+        if path.is_file():
+            if path.suffix.lower() == ".eml":
+                email_files.append(path)
+            else:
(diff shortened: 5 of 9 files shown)


bmerkle:
(typeagent) PS C:\work\microsoft\typeagent-py> .\make.bat test
Running unit tests...
================================================================= test session starts =================================================================
platform win32 -- Python 3.13.5, pytest-8.4.2, pluggy-1.6.0
rootdir: C:\work\microsoft\typeagent-py
configfile: pyproject.toml
testpaths: test
plugins: anyio-4.11.0, logfire-4.14.2, asyncio-1.2.0, mock-3.15.1
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 358 items                                                                                                                                    

test\test_add_messages_with_indexing.py ...                                                                                                      [  0%]
test\test_auth.py .......                                                                                                                        [  2%]
test\test_collections.py ...........................                                                                                             [ 10%]
test\test_conversation_metadata.py ....................                                                                                          [ 15%]
test\test_demo.py .                                                                                                                              [ 16%]
test\test_embedding_consistency.py ...                                                                                                           [ 17%]
test\test_embeddings.py ...........                                                                                                              [ 20%]
test\test_factory.py ...                                                                                                                         [ 20%]
test\test_incremental_index.py ..                                                                                                                [ 21%]
test\test_interfaces.py ......................                                                                                                   [ 27%]
test\test_knowledge.py ....                                                                                                                      [ 28%]
test\test_kplib.py ......                                                                                                                        [ 30%] 
test\test_mcp_server.py ss                                                                                                                       [ 31%]
test\test_message_text_index_population.py .                                                                                                     [ 31%]
test\test_message_text_index_serialization.py ..                                                                                                 [ 31%] 
test\test_messageindex.py ....s...                                                                                                               [ 34%]
test\test_online.py .                                                                                                                            [ 34%]
test\test_podcast_incremental.py ..                                                                                                              [ 34%]
test\test_podcasts.py .                                                                                                                          [ 35%]
test\test_property_index_population.py .                                                                                                         [ 35%]
test\test_propindex.py ............                                                                                                              [ 38%]
test\test_query.py ..................................                                                                                            [ 48%]
test\test_query_method.py ..                                                                                                                     [ 48%]
test\test_related_terms_fast.py .                                                                                                                [ 49%]
test\test_related_terms_index_population.py .                                                                                                    [ 49%]
test\test_reltermsindex.py .............                                                                                                         [ 53%]
test\test_searchlib.py .......................................................                                                                   [ 68%]
test\test_secindex.py ...                                                                                                                        [ 69%] 
test\test_secindex_storage_integration.py .                                                                                                      [ 69%]
test\test_semrefindex.py ......................                                                                                                  [ 75%]
test\test_serialization.py ........                                                                                                              [ 77%]
test\test_sqlite_indexes.py .....................                                                                                                [ 83%]
test\test_sqlitestore.py .....                                                                                                                   [ 85%]
test\test_storage_providers_unified.py ..............................                                                                            [ 93%]
test\test_timestampindex.py .                                                                                                                    [ 93%] 
test\test_transcripts.py .......                                                                                                                 [ 95%]
test\test_utils.py ....                                                                                                                          [ 96%]
test\test_vectorbase.py ...........                                                                                                              [100%]

====================================================== 355 passed, 3 skipped in 64.52s (0:01:04) ======================================================

gvanrossum:
Since you touched so many files, there was a conflict with a PR I merged in the last 30 minutes. I fixed the conflict (it was trivial) and this has the side effect of running the tests under my identity, so they will actually run. (I think we need to adjust something in the way we get the auth tokens.)

gvanrossum:
Also, I'm not sure how soon I'm going to get to the review -- that's a *lot* of changes, mostly simple, and I don't want to fall asleep while double checking them all.

bmerkle:
sorry for the PR conflict. The changes overlapped in terms of timing.

For review the simple changes, in future a agent would be cool 😸.
But for now we could have a pre-commit hook which formats code and checks for inline imports (to avoid) them.

OTOH it would be cool to merge the PR soon, to avoid following PR conflicts, and most of the stuff is only moved/sorted  imports

gmail/gmail_dump.py
```diff
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
 
 import argparse
+import time
```
gvanrossum, line 5: The way I like my imports sorted (by full module name), 'time' follows 'pathlib'.

test/fixtures.py
```diff
@@ -1,26 +1,25 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from collections.abc import AsyncGenerator, Iterator
 import os
 import tempfile
+from collections.abc import AsyncGenerator, Iterator
```
gvanrossum, line 6: Similar here.

test/fixtures.py
```diff
@@ -1,26 +1,25 @@
...
 from typing import Any
 
 import pytest
 import pytest_asyncio
-
+import tiktoken
```
gvanrossum, line 11: And here.

test/fixtures.py
```diff
@@ -1,26 +1,25 @@
...
 from openai.types.create_embedding_response import CreateEmbeddingResponse, Usage
 from openai.types.embedding import Embedding
-import tiktoken
 
 from typeagent.aitools import utils
-from typeagent.aitools.embeddings import AsyncEmbeddingModel, TEST_MODEL_NAME
+from typeagent.aitools.embeddings import TEST_MODEL_NAME, AsyncEmbeddingModel
```
gvanrossum, line 16: Leave the ordering alone. Imported things go alphabetically.

test/test_add_messages_with_indexing.py
```diff
@@ -8,7 +8,7 @@
 
 import pytest
 
-from typeagent.aitools.embeddings import AsyncEmbeddingModel, TEST_MODEL_NAME
+from typeagent.aitools.embeddings import TEST_MODEL_NAME, AsyncEmbeddingModel
```
gvanrossum, line 11: Again on both count -- why change this file, and why reorder the items here.

test/fixtures.py
```diff
@@ -1,26 +1,25 @@
 # Copyright (c) Microsoft Corporation.
```
gvanrossum, line 1: Why change this file at all? I it has no inline imports.

test/test_message_text_index_serialization.py
```diff
@@ -3,19 +3,23 @@
 
 """Test for MessageTextIndex serialization to ensure it's no longer a no-op."""
 
-import pytest
 import sqlite3
 
 import numpy as np
+import pytest
```
gvanrossum, line 9: The 3rd party modules should be in one block. But fixtures is local and needs to be separated by a blank line.

test/test_online.py
```diff
@@ -2,7 +2,6 @@
 # Licensed under the MIT License.
 
 import pytest
-
```
gvanrossum, line 5: Put that blank line back please.

test/test_searchlib.py
```diff
@@ -110,8 +113,8 @@ def test_create_and_term_group(self):
 
         assert group.boolean_op == "and"
         assert len(group.terms) == 2
-        assert group.terms[0] == term1
-        assert group.terms[1] == term2
+        assert pydantic_dataclass_to_dict(group.terms[0]) == pydantic_dataclass_to_dict(term1)
+        assert pydantic_dataclass_to_dict(group.terms[1]) == pydantic_dataclass_to_dict(term2)
```
gvanrossum: Curious why this was needed. Doesn't pydantic's dataclass create an `__eq__` method etc.?

tmp_debug_tag.py
```diff
@@ -0,0 +1,14 @@
+"""Legacy debug script placeholder.
```
gvanrossum, line 1: This file probably shouldn't be imported. If it should, it needs a copyright header.

tools/add_copyright.py
```diff
@@ -18,7 +18,6 @@
 from pathlib import Path
 from typing import List, Tuple
 
-
```
gvanrossum, line 21: Please don't mess with this spacing.

tools/get_keys.py
```diff
@@ -13,10 +13,11 @@
 from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple
 
+import colorama
+
 # Azure SDK imports
```
gvanrossum, line 18: I'd rather change this comment than move the imports.

tools/vizcmp.py
```diff
@@ -8,7 +8,8 @@
 import statistics
 import sys
 
-from colorama import init as colorama_init, Back, Fore, Style
+from colorama import Back, Fore, Style
+from colorama import init as colorama_init
```
gvanrossum, line 12: Why this change?

typeagent/aitools/auth.py
```diff
@@ -2,8 +2,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from dataclasses import dataclass
 import time
+from dataclasses import dataclass
```
gvanrossum, line 6: Original was better...

typeagent/knowpro/convsettings.py
```diff
@@ -77,8 +81,8 @@ def storage_provider(self, value: IStorageProvider) -> None:
     async def get_storage_provider(self) -> IStorageProvider:
         """Get or create the storage provider asynchronously."""
         if self._storage_provider is None:
+            # Import here to avoid circular import
             from ..storage.memory import MemoryStorageProvider
-
```
gvanrossum, line 81: Doesn't 'make format' put this blank line back?

gvanrossum (commented):
Flight's leaving.

typeagent/knowpro/searchlib.py
```diff
@@ -6,8 +6,30 @@
 Functions that help with creating search and property terms
 """
 
-from typing import cast
+import dataclasses
+from typing import Any, cast
+
+
+def pydantic_dataclass_to_dict(obj: Any) -> Any:
```
gvanrossum, line 13: Why do we need this? It seems significant but I fail to see why.

typeagent/storage/__init__.py
```diff
@@ -5,14 +5,14 @@
 
 # Import from new organized structure
 from .memory import (
-    MemoryStorageProvider,
-    MemoryMessageCollection,
-    MemorySemanticRefCollection,
+                     MemoryMessageCollection,
+                     MemorySemanticRefCollection,
+                     MemoryStorageProvider,
```
gvanrossum: Bad formatting. Use 'make format'

typeagent/storage/memory/provider.py
```diff
@@ -74,7 +72,6 @@ async def __aexit__(
         exc_tb: object,
     ) -> None:
         """Exit transaction context. No-op for in-memory storage."""
-        pass
```
gvanrossum, line 77: I like to preserve such 'pass' lines.

typeagent/storage/sqlite/collections.py
```diff
@@ -7,9 +7,8 @@
 import sqlite3
 import typing
 
+from ...knowpro import interfaces, serialization
 from .schema import ShreddedMessage, ShreddedSemanticRef
```
gvanrossum, line 11: Correct! My guiding principle here is that the "closest" import comes last.

typeagent/storage/sqlite/provider.py
```diff
@@ -215,7 +210,6 @@ def _check_embedding_consistency(self) -> None:
         Raises:
             ValueError: If embeddings in the database don't match the expected size.
         """
-        from .schema import deserialize_embedding
```
gvanrossum, line 213: If you delete the last online import, also delete the blank line (which was imposed by the formatter).

bmerkle:
@gvanrossum question: can we have a simple, low-threshold communcation channel (like chat, discord, ms-teams, or linked-in chat) to brainstrom or discuss in advance things ? Then i might explain beforehand and we can easily chat if it make sense. :-) 

in the PR here I used ruff instead of isort and format. It might sort thing differently than isort so this change caused some noise or disagreement about sorting and formatting i think. Appended is a short comparison of ruff vs. isort and format.

bmerkle:
✅ What Ruff brings — and where it may “beat” isort

Ruff is a fast, all-in-one linter + formatter + import-sorter — it tries to replace not only isort, but also tools like linters, formatters, and more in one go. 

If you enable the import-sorting rules (rule group “I”), Ruff can sort imports — so you don’t need a separate tool just for imports. 
Medium


For large code bases or in continuous-integration / pre-commit settings, Ruff tends to be much faster than combining multiple tools (formatter + linter + isort) because it’s implemented in Rust and optimized for performance. 

For convenience: with Ruff you can have linting + formatting + import handling in a single command or setup, reducing the number of dependencies or tools you need to maintain. 

So if you want a unified toolchain, care about speed/performance, and are okay with potentially slightly different import-sorting behavior — Ruff is a compelling modern choice.

⚠️ Where isort still shines (or where Ruff may fall short)

Import sorting in Ruff is “intended to be near-equivalent” to isort (especially if you configure isort with certain profiles), but there are known differences — e.g. in how aliased imports are grouped, or how inline comments are handled. 

Ruff’s import-sorting capabilities are “good enough” for many, but if you rely on very precise import ordering/grouping rules, or need maximum control over formatting decisions, isort might be more predictable and configurable.

Because Ruff is broader in scope (linter + formatter + sorter), if something goes wrong — configuration issues, conflicts with formatting — it may be harder to isolate and debug than a dedicated import sorter.

Some users report edge cases or inconsistencies when mixing Ruff’s sorting + formatting + other tools (especially if also using other formatters or import tools).

bmerkle:
i did not run `make format check` , and the refactoring broke some typechecking tests.
sorry i will rework the PR.

test/fixtures.py
bmerkle: ruff formats imports according to the PEP 8 import ordering rules, which generally means:
- first import ...
- then from ... import ...

Ruff’s import sorter (ruff --fix, rule I001, internally inspired by isort) enforces:
Standard library imports
Third-party imports
Local application imports

bmerkle:
@gvanrossum okay, much of the noise was caused by the reformatting of imports at the top of the file.
I did not only fix inline imports, but at the same time rearranged the imports because i had to touch them anyway.
Via ruff i did a automatic reformatting, but it should conform to PEP 8 and be consistent because it is automatic.
So hopefully this explains the modifications regarding imports. I looked through all files and they LGTM.


ruff formats imports according to the PEP 8 import ordering rules, which generally means:
- first import ...
- then from ... import ...

Ruff’s import sorter (ruff --fix, rule I001, internally inspired by isort) enforces:
- Standard library imports
- Third-party imports
- Local application imports

tools/add_copyright.py
bmerkle: sorry, I missed make format

bmerkle:
### relevant changes: 
Added dataclasses.py to wrap pydantic’s decorator with dataclass_transform + overloads so pyright understands generated initializers and fields; 
updated all knowpro modules, tests, and related utilities to import the wrapper instead of pydantic.dataclasses.dataclass.

Adjusted dependent modules (interfaces.py, kplib.py, search.py, search_query_schema.py, date_time_schema.py, answer_response_schema.py, universal_message.py, email_message.py, and several tests) to use the new wrapper while preserving import ordering and existing behavior.

### note: 
all other changes a import and formatting related modifications (which should be ok)
`make format, check, test` is running clean

test/test_searchlib.py
bmerkle: If you use Pydantic dataclasses, you get the normal Python dataclass dunder methods, including __eq__.
If you use Pydantic models, equality is handled differently, there is no generation of __eq__ etc. 

you are right that we are currently using dataclasses and not models, but if we would switch to models, the new code above would also work (so in this case it is more generic and prepared for changes in the future) .

gvanrossum:
> chat

I'm on Discord but I don't want to reveal my handle here and I have heavy privacy barriers around it. If you email me privately at guido@python.org we can arrange something.

KRRT7:
Generally in vibe-coded projects there tend to be a lot of inline imports because LLMs are bad at writing code properly and tend to hack around things by adding inline imports vs structuring and importing the right way the first time around.

My suggestion would be to reset against main, implement the necessary changes, and then run `make format` only, strictly resolving the inline imports / implementing the refactor. This way you'll have an easy-to-review PR for this specific concern.

Later on you could make a separate PR switching over to ruff and avoiding the 101 files diff in *this* PR and have it happen in that subsequent PR

Another thing: instead of asking LLMs about the differences between isort and ruff import sorting, I'd highly suggest reading over https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort as it's up to date.

bmerkle:
> My suggestion would be to reset against main, implement the necessary changes, and then run `make format` only, strictly resolving the inline imports / implementing the refactor. This way you'll have an easy-to-review PR for this specific concern.
> 
> Later on you could make a separate PR switching over to ruff and avoiding the 101 files diff in _this_ PR and have it happen in that subsequent PR

Hi @KRRT7 thanks for your comments.
regarding the PR i am open to reset against main, and do it again, but i think we will end up with the same number of modification. 

Refactoring the inline import is a kind of mass operations when it affects several files.
I also automatically rearranged the import and from statements of the beginning of the files.
While they follow the accepted PEP style and have been done consistent and automatically with ruff, they have raised some discussion (alphabetical sort in the list, etc).
 
The interesting part of this PR is the removal of circular dependencies. I can outline and explain this to better understand the PR.

The rest of the PR is pretty trivial as it concerns only the above mentioned import and inline rearrangement and it by nature a mass change. I reviewed this in VSCode and you have only to really check the beginning sections of the files. And the inline imports which have been pulled up to the top of the file.

You could argue that i have done all 3 things (remove circular dependencies, remove inline import, reformat imports) in one step, but i think it made sense.

Please lets me know what you think. I can resubmit 3 PR for each step, but if we reformat the imports, number of modifications we be the same.

gvanrossum:
I strongly prefer to undo all the import reformats for this PR, if possible. The other two are intertwined and should probably be kept together.

bmerkle:
OK i will create a new PR.

bmerkle:
please cancel/close this PR as it is too large and covers too much different aspects.
I will create 3 separate, smaller PR for each aspect:
- reduce curicular dependencies https://github.com/microsoft/typeagent-py/pull/118
- fix inline imports (still TODO)
- fix import layout (still TODO)

PR #115, opened 2025-12-04
feat(storage): Implement `SqliteRelatedTermsIndex.serialize()` method

iamrajiv:
I noticed the NotImplementedError in the serialize methods while exploring the codebase. I implemented the missing serialize() functionality for:

- SqliteRelatedTermsFuzzy.serialize()
- SqliteRelatedTermsIndex.serialize()

### DIFF
--- a/tools/query.py
+++ b/tools/query.py
@@ -1,14 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-__version__ = "0.2"
+__version__ = "0.3"
 
 ### Imports ###
 
 import argparse
 import asyncio
 from collections.abc import Mapping
-from dataclasses import dataclass
+from dataclasses import dataclass, field, replace
 import difflib
 import json
 import os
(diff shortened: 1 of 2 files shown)


test/test_sqlite_indexes.py
```diff
@@ -401,6 +434,94 @@ async def test_combined_index_basic(
...
+        assert text_embedding_data is not None
+        assert "textItems" in text_embedding_data
+        assert len(text_embedding_data["textItems"]) == 3
+        assert "chess" in text_embedding_data["textItems"]
+        assert "grandmaster" in text_embedding_data["textItems"]
+        assert "embeddings" in text_embedding_data
+
+        # Create a fresh database and index to test deserialization
+        import tempfile
+        import os
```
gvanrossum: These two imports should be at the top level.

test/test_sqlite_indexes.py
```diff
@@ -401,6 +434,94 @@ async def test_combined_index_basic(
...
+            fresh_db = sqlite3.connect(fresh_db_path)
+            init_db_schema(fresh_db)
+
+            fresh_index = SqliteRelatedTermsIndex(fresh_db, embedding_settings)
+
+            # Deserialize into fresh index
+            await fresh_index.deserialize(data)
+
+            # Verify alias data was restored
+            alias_results = await fresh_index.aliases.lookup_term("AI")
```
gvanrossum: Do you want to rename this variable to `ai_results` for consistency with `ml_results` below?

test/test_sqlite_indexes.py
```diff
@@ -401,6 +434,94 @@ async def test_combined_index_basic(
...
+            assert ml_results is not None
+            assert len(ml_results) == 1
+            assert ml_results[0].text == "machine learning"
+
+            # Verify fuzzy index data was restored
+            fresh_fuzzy = fresh_index.fuzzy_index
+            assert fresh_fuzzy is not None
+            assert await fresh_fuzzy.size() == 3
+
+            # Use concrete type to access get_terms (not in interface)
```
gvanrossum, line 509: I wonder if `get_terms()` *should* be added to the interface. (Probably not in this PR unless it's a one-liner -- which it isn't since you'd have to implement it for the memory provider too.)

gvanrossum:
@iamrajiv If you're around could you address my nits? Then we can merge it.

test/test_sqlite_indexes.py
iamrajiv: sure done

test/test_sqlite_indexes.py
iamrajiv: done

test/test_sqlite_indexes.py
iamrajiv: @gvanrossum I agree get_terms() could be useful in the interface. I'll keep it out of this PR to keep the scope small, but happy to open a follow-up issue/PR to add it to the interface and implement it for both providers if that would be helpful.

iamrajiv:
Thank you for the review, @gvanrossum. I made the changes you suggested. Could you please review it again?

gvanrossum (approved):
Perfect! Merging now.

PR #118, opened 2025-12-09
Wrap pydantic.dataclass; split interfaces.py into parts

bmerkle:
* Add knowpro/dataclasses.py, which contains a wrapper for pydantic.dataclasses.dataclass that preserves the original class's type -- and change all the imports to use this instead of importing pydantic.dataclasses.dataclass directly.
* Split interfaces into five thematic parts.

### DIFF
--- a/typeagent/knowpro/interfaces.py
+++ b/typeagent/knowpro/interfaces.py
@@ -4,8 +4,6 @@
 
 from __future__ import annotations
 
-from typing import Final
-
 from .interfaces_core import *
 from .interfaces_indexes import *
 from .interfaces_search import *
@@ -18,4 +16,5 @@ from .interfaces_search import __all__ as _search_all
 from .interfaces_serialization import __all__ as _serialization_all
 from .interfaces_storage import __all__ as _storage_all
 
+# pyright: reportUnsupportedDunderAll=false
 __all__ = _core_all + _indexes_all + _search_all + _serialization_all + _storage_all


gvanrossum:
Interesting. Can you explain why you made these changes? Everything passes without them as well, so why are they necessary? Let me first summarize what you did:

1. Added knowpro/datackasses.py, containing a wrapper for `pydantic.dataclasses.dataclass` that has a better type signature; and import it in some places instead of the latter. (Did you replace *all* imports of the latter?)
2. Added knowpro/types.py, containing a variant of `ConversationDataWithIndexes` that uses `Any` for some of the container fields, and a version of `SearchTermGroupTypes` that's just `Any`.  And then you import these instead of the originals from `interfaces`. (Again, do you use these fake ones everywhere?)

How did you determine that precisely these changes were needed, and is there something I can do that shows some check or test failing without these changes? I'm a bit uncomfortable with using the fake types from types.py with all those `Any`s, and I worry that the fake `ConversationDataWithIndexes` might be instantiated and then fail an `isinstance` check against the real class.

I have an open mind, but I can't read your mind regarding the motivation and scope of these changes. Thanks in advance for your explanation!

bmerkle:
> 1. Added knowpro/datackasses.py, containing a wrapper for `pydantic.dataclasses.dataclass` that has a better type signature; and import it in some places instead of the latter. (Did you replace _all_ imports of the latter?)

it is a small functionally equivalent wrapper around `pydantic.dataclasses.dataclass`.  The idea was to reduce the circular dependencies and have a small wrapper around the pydantic lib, so we could e.g. exchange it also with another lib/framework in future and/or have a central place where we can react to pydantic api changes. I refactored all imports where it was used.
I checked again and found 4 places we still import pydantic directly (test_utils.py, email_memory.py, podcast.py and transcript.py), but AFAIK the imports are not use, so can be removed safely.


> 2\. Added knowpro/types.py, containing a variant of `ConversationDataWithIndexes` that uses `Any` for some of the container fields, and a version of `SearchTermGroupTypes` that's just `Any`.  And then you import these instead of the originals from `interfaces`. (Again, do you use these fake ones everywhere?)

ok. I see your point and it is valid. I cut the dependencies and cirucular imports, but we loose the typesafety which is not good.
Loosing the typesafety and contracts is not a good idea. So types.py as it is implemented now it a bad idea.

OTOH interfaces.py is quite large and contains different aspects AFAIS:
(1) conversation + message data structures and TypedDicts (serialization), 
(2) indexing protocols and helper dataclasses, 
(3) storage-provider interfaces, 
(4) search/query models, and 
(5) knowledge extraction helpers

So would it make sense to incrementally split up intefaces.py into these different aspects and create own files for them ?

gvanrossum:
Yeah, splitting interfaces.py makes sense to me. It was originally one file because the original author of the code (in TypeScript) wanted all public interfaces together, but I think that hasn't worked out as well as we had hoped. I'm okay exporting everything through interfaces but defining the different categories in different files, which interfaces then imports * from.

Regarding data classes, I don't really like the idea of complicating code for the sake of something that may but probably won't happen in the future (look up YAGNI).

I was hoping that the extra type-safety would show somewhere. E.g. areas that are untyped without the change becoming typed with it? Did you see any of that?

typeagent/knowpro/types.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Shared type helpers used to break circular imports in knowpro."""
+
+from typing import Any, Generic, NotRequired, TypedDict, TypeVar
+
+TMessageData = TypeVar("TMessageData")
```
gvanrossum: Let's please use the modern notation for generics as used elsewhere in the codebase.

bmerkle:
Ok. I will slowly split up interfaces.py and make a first suggestion (without touching other Code). We then can proceed from there If you agree on the split up Iteration.

Regarding typesafety my apporach was wrong direction. I have to investigate. I will be on the Road tomorrow and probably continue on Tue.

gvanrossum:
I would start by listing the different categories of definitions present in interfaces.py.

I think that perhaps the first category to move out could be the serialization results named like FooData?

The storage providers also deserve to be split off.

Separately, have you made an analysis of the circular imports? I'd do that before doing more about them.

Regarding the pydantic.dataclass wrapper, I wonder if there are actually noticeable effects -- I sometimes notice type errors in VScode that might be explained this way. Maybe we should pressure pydantic to fix this though?

gvanrossum:
Oops, I edited that comment extensively. Please be sure to read the version on GitHub.com.

bmerkle:
okay a few questions before i commit and push the changes:

- i have removed types.py again because it made no sense as discussed

- shall we keep the knowpro.dataclasses.py as wrapper for pydantic.dataclasses.dataclass ? 
For the moment i think it does not really hurt to have this small wrapper, but if you say YAGNI, I can also remove it again

- I have split up the content of interfaces.py in 5 separate files (as discussed above in https://github.com/microsoft/typeagent-py/pull/118#issuecomment-3679390993).
Everything is backward compatible because interfaces.py import all the new files so no client changes are necessary.
But the separate concerns are now in separate files, so semantic and modularization is better.

- during move of code i found two things I would like to sync with you: (old and new version)

- I would use python docstrings instead of prepending # comments
```
# Return text ranges in the given date range.
class ITimestampToTextRangeIndex(Protocol):
```

```
class ITimestampToTextRangeIndex(Protocol):
    """Return text ranges over a date range."""

```
- I think the following might be a bug/feature. I have removed the " " , because then it is evaluated eagerly and it can be detected by pydantic, so this change should improve typesafety and avoid circular imports....

```
async def serialize(self) -> "TermToRelatedTermsData": ...
```

```
 async def serialize(self) -> TermToRelatedTermsData: ...
```

typeagent/knowpro/types.py
bmerkle: I have added also instructions in AGENTS.md to assert the behavior

gvanrossum:
> shall we keep the knowpro.dataclasses.py as wrapper for pydantic.dataclasses.dataclass ?

Let's drop it. I thought it resulted in fewer red squiggly lines, but that disappeared after restarting VS Code.

> I would use python docstrings instead of prepending # comments

Agreed.

> I think the following might be a bug/feature. I have removed the " " , because then it is evaluated eagerly and it can be detected by pydantic, so this change should improve typesafety and avoid circular imports....

Whenever you see such "" around a type annotation it means that I couldn't come up with a decent ordering where no class was used before it was defined. If removing the "" doesn't cause more errors during `make check test` then they were redundant and should be removed. So it sounds like you should remove them.

bmerkle:
Okay so commit #6d6977e should be okay to be reviewed by you.

- Split up interfaces.py in to separate concerns. 
We have 5 new files and modification is backward compatible

- reviewed typeagent.knowpro.dataclasses again
I tried back and forth with this file, but i suggest that we stay with it as a thin wrapper around the pydantic library.
Without this thin layer, my experience was, that Pydantic’s dataclass decorator returns a dynamically generated class that type checkers treat as “opaque,” so Pyright can’t see any of the fields the decorator added and reports tons of errors. (I though it would be a pyright issue and updated to 1.1.407 but that has a regression bug)

This small wrapper re-exposes Pydantic’s decorator, and declares overloads so Pyright/Mypy knows the decorator returns the same class type they passed in. With it in place, every KnowPro dataclass (messages, search terms, schema objects, etc.) has a statically visible constructor and attributes. Without it, we have to adapt that guard in each module of knowpro, which is more an duplicated work. I have updated dataclasses.py to conform to PEP 695 syntax. (as we rely on python >= 3.12)

I had some issues with pyright version 1.1.407 (which is broken) and the warning it always emits that there is a new version.
Tried to silence it (suppress the warning) but I had no success (I set PYRIGHT_PYTHON_FORCE_VERSION=1.1.406) it ignores it.
npm is really a mess with package handling and update behaviour....

gvanrossum:
Okay, I see now. We should also file a bug with Pydantic.

test/test_mcp_server.py
```diff
@@ -15,6 +15,18 @@
 
 from fixtures import really_needs_auth
 
+if TYPE_CHECKING:
+    from openai.types.chat import ChatCompletionMessageParam
+else:  # pragma: no cover - optional dependency
+    try:
+        from openai.types.chat import ChatCompletionMessageParam
+    except ImportError:
+        ChatCompletionMessageParam = dict[str, Any]  # type: ignore[assignment]
```
gvanrossum: Under what circumstances would we get this import error??

test/test_mcp_server.py
```diff
@@ -15,6 +15,18 @@
...
+
+pytestmark = pytest.mark.skip(
+    reason="mcp server tests require interactive dependencies; skipping for now"
+)
```
gvanrossum: Is this still used?

test/test_mcp_server.py
bmerkle: not relevant anymore, 
let me push 2 files again which had old workarounds from the initial mcp problems we debugged.

test/test_mcp_server.py
bmerkle: not relevant anymore, 
let me push 2 files again which had old workarounds from the initial mcp problems we debugged.

bmerkle:
git #cf76889 is now cleaned up version. (removed old superfluous stuff)

.gitignore
```diff
@@ -21,6 +21,9 @@ __pycache__
 /*_data.json
 /*_embeddings.bin
 
+# pytest configuration
+pytest.ini
+
```
gvanrossum, line 26: Do we need this? I imagine if we had a pytest.ini file we'd probably want to have it under version control (i.e., git).

gvanrossum: (My comment seemed lost, or I am confused by GitHub's UI.)

Why do we need this? I'd imagine *if* we had a pytest.ini file we'd want it in git.

typeagent/knowpro/interfaces.py
```diff
@@ -1,911 +1,82 @@
...
+
+from __future__ import annotations
+
+from .interfaces_core import *  # noqa: F401,F403
+from .interfaces_indexes import *  # noqa: F401,F403
+from .interfaces_search import *  # noqa: F401,F403
+from .interfaces_serialization import *  # noqa: F401,F403
+from .interfaces_storage import *  # noqa: F401,F403
+
+__all__ = [
```
gvanrossum: It feels brittle to repeat `__all__` here. Could we perhaps import each submodule's `__all__` and concatenate them? E.g.

```py
from blah1 import __all__ as _all1
from blah2 import __all__ as _all2
...

__all__ = _all1 + _all2 + ...
```

typeagent/knowpro/interfaces.py
```diff
@@ -1,911 +1,82 @@
...
-    "topic",  # topic
+"""Aggregated knowpro interfaces for backwards compatibility."""
+
+from __future__ import annotations
+
+from .interfaces_core import *  # noqa: F401,F403
+from .interfaces_indexes import *  # noqa: F401,F403
+from .interfaces_search import *  # noqa: F401,F403
+from .interfaces_serialization import *  # noqa: F401,F403
+from .interfaces_storage import *  # noqa: F401,F403
```
gvanrossum: What tool requires the `# noqa` comments? Not any of the tools we're currently using in the repo or CI, I believe.

Or is this part of switching to ruff instead of black? I'm not yet convinced that we need that. It certainly seems we'd need to think about that separately. (It does appear to find a lot of redundant imports and variable assignments.)

bmerkle: yep, it is for ruff. 
I can delete the comments until we probably decide to go (or not go) with ruff

gvanrossum: Okay please do.

.gitignore
bmerkle: Hm, I could make a pytest.local.ini to contain my local config which should not be version controled.
We have then to invoke  pytest -c pytest.ini -c pytest.local.ini in CI as we have to opt-in the file.

bmerkle:
> Did you make any changes to the split-out files? I'm not going to review all of those.

`async def serialize(self) -> "TermToRelatedTermsData": ...`
changed to
`async def serialize(self) -> TermToRelatedTermsData: ...`
as discussed above

typeagent/knowpro/interfaces.py
```diff
@@ -4,79 +4,18 @@
 
 from __future__ import annotations
 
-from .interfaces_core import *  # noqa: F401,F403
-from .interfaces_indexes import *  # noqa: F401,F403
-from .interfaces_search import *  # noqa: F401,F403
-from .interfaces_serialization import *  # noqa: F401,F403
-from .interfaces_storage import *  # noqa: F401,F403
+from typing import Final
```
gvanrossum: This seems a spurious import?

gvanrossum (changes requested):
Just one more nit and then I'm ready to approve.

typeagent/knowpro/interfaces.py
bmerkle: fixed and suppressed pyright warning. Should now be ready to approve :-)

gvanrossum:
@bmerkle Thanks for your patience!

bmerkle:
Thanks for your help and feedback. Appreciate a lot :-)

PR #117, opened 2025-12-04
feat: Add intelligent conversation history for pronoun/reference resolution

iamrajiv:
Fixes: #58

### How It Works

```
                        BEFORE (No History)
+-------------------------------------------------------------------------+
|  User: "What did Adrian say about spiders?"                             |
|  Answer: "Adrian mentioned spiders have eight legs..."                  |
|                                                                         |
|  User: "What else did he say about it?"                                 |
|         |                                                               |
|         v                                                               |
|  LLM: "he"=? "it"=? ---> [X] Cannot resolve references                  |
+-------------------------------------------------------------------------+

                        AFTER (With History)
+-------------------------------------------------------------------------+
|                                                                         |
|  +---------------+    +------------------+    +-----------------------+ |
|  | User Query    |--->| History Context  |--->| LLM Query Translation | |
|  |               |    | (last 3-5 Q&A)   |    |                       | |
|  | "What else    |    |                  |    | Prompt includes:      | |
|  |  did he say   |    | Q1: What did     |    | - User query          | |
|  |  about it?"   |    |     Adrian say   |    | - Recent Q&A pairs    | |
|  |               |    |     about        |    |                       | |
|  +---------------+    |     spiders?     |    | LLM resolves:         | |
|                       | A1: Adrian       |    | "he" -> Adrian        | |
|                       |     mentioned    |    | "it" -> spiders       | |
|                       |     spiders...   |    |                       | |
|                       +------------------+    +-----------+-----------+ |
|                                                           |             |
|                                                           v             |
|                                               +-----------------------+ |
|                                               | Structured Search     | |
|                                               | Query (JSON)          | |
|                                               |                       | |
|                                               | entities: ["Adrian"]  | |
|                                               | topics: ["spiders"]   | |
|                                               +-----------------------+ |
|                                                           |             |
|                                                           v             |
|                                                    Correct Answer       |
+-------------------------------------------------------------------------+
```

### Data Flow

```
+-------------+     +--------------------+     +---------------------+
|             |     |                    |     |                     |
| User Query  |---->| process_query()    |---->| search_conversation |
|             |     |                    |     | _with_language()    |
+-------------+     | 1. Get history     |     |                     |
                    |    prompt section  |     | model_instructions  |
                    | 2. Inject into     |     | contains history    |
                    |    lang_options    |     | context             |
                    +--------------------+     +----------+----------+
                                                          |
                    +--------------------+                v
                    |                    |     +---------------------+
                    | ConversationHistory|<----| search_query_from   |
[...]

PR #120, opened 2025-12-23
Update uv.lock with new typeagent version

gvanrossum:
We should really do this in tools/release.py, but we didn't last time, so here goes. This finishes the release of 0.3.3. :-)


gvanrossum:
Never mind this was included in #117.

Issue #125, opened 2025-12-28
migrate to python standard project layout


bmerkle:
Before we make further improvement, i would suggest to tackle the following two small issues from TODO.md

### P3 - Low Priority
- **[P3, small]** Move `typeagent` into `src/`
- **[P3, tiny]** Move `test/` to `tests/`

The idea would be to migrate to a python standard project layout:
- no semantic code changes
- follow python standard project layout (for ML/AI projects)
- if necessary adopt config files, scripts etc to 
- preserve history via git mv commands

## migration for typeagent-py
typeagent-py/
├─ pyproject.toml (etc)
├─ README.md (etc)
├─ Makefile (etc)
├─ data/
│  ├─ demo.db (etc)
├─ docs/
│  └─ spec/
├─ src/
│  └─ typeagent/
│     ├─ __init__.py
│     ├─ dataset.py
│     ├─ model.py
│     └─ train.py
├─ examples/
│  └─ demo/
│  └─ examples/
├─ tests/
│  └─ test_model.py
│  └─ testdata
├─ tools/
│     └─ gmail/

## migration steps

- [x] introduce src 125 src-create
mkdir src
git add src
git mv typeagent src/typeagent

- [X] move test 125-tests
git mv test tests
git mv testdata tests/testdata

- [x] consolidate examples 125-examples
git mv demo examples/demo
git mv samples examples/samples
git mv gmail tools/gmail
- afterwards we decide which are outdated or should be updated/removed/new added

- [x] move spec-to-docs 125-spec
git mv spec docs/spec

## questions
- shoud we move *.db to data/ (and adopt .gitignore)
- should we move some of the .md files to doc/ (which one)
- we should have a .env-template file (as example for .env) only containing names of the API-KEYS...

bmerkle:
I would prepare a PR, as described above, but would need your opinion @gvanrossum about layout and the open questions.
Feel free to edit also the initial issue description. Thanks you so much.

gvanrossum:
Leave tools where it is. Don't move the .md files yet. Don't do the samples. I propose separate PRs for src and tests and not changing anything else in this round.

gvanrossum:
Do update TODO.

gvanrossum:
Don't move dist -- IIRC it's the default destination used by uv's default build tool.

bmerkle:
> Do update TODO.

do you mean i should move it to docs or update only the content ?
I will update TODO.md content once we are through this issue.

[...]

PR #127, opened 2025-12-28
created src and move typeagent to src

bmerkle:
- create src
- move typeagent to src
- adapt config files

https://github.com/microsoft/typeagent-py/issues/125


pyproject.toml
```diff
@@ -64,8 +64,8 @@ GitHub = "https://github.com/microsoft/typeagent-py"
 Documentation = "https://github.com/microsoft/typeagent-py/tree/main/docs/README.md"
 
 [tool.uv.build-backend]
-module-root = ""
+module-root = "src"
 
 [tool.pytest.ini_options]
 asyncio_default_fixture_loop_scope = "function"
-testpaths = ["test"]
+testpaths = ["test"]
```
gvanrossum: Looks like you accidentally removed the final newline from this file.

make.bat
```diff
@@ -26,13 +26,13 @@ goto help
...
 echo Formatting code...
-.venv\Scripts\black typeagent test tools gmail demo
+.venv\Scripts\black src\typeagent test tools gmail demo
 goto end
 
 :check
 if not exist ".venv\" call make.bat venv
 echo Running type checks...
-.venv\Scripts\pyright --pythonpath .venv\Scripts\python typeagent test tools gmail demo
+.venv\Scripts\pyright --pythonpath .venv\Scripts\python src\typeagent test tools gmail demo
```
gvanrossum:
```suggestion
.venv\Scripts\pyright --pythonpath .venv\Scripts\python src test tools gmail demo
```

make.bat
```diff
@@ -26,13 +26,13 @@ goto help
 :format
 if not exist ".venv\" call make.bat venv
 echo Formatting code...
-.venv\Scripts\black typeagent test tools gmail demo
+.venv\Scripts\black src\typeagent test tools gmail demo
```
gvanrossum:
```suggestion
.venv\Scripts\black src test tools gmail demo
```

gvanrossum (approved):
Just two tiny suggestions and one formatting nit. I will merge after you fix those.

gvanrossum:
Oh, and you need to merge the conflict caused by me merging my unused-xxx fix first (Rob was okay reviewing it).

gvanrossum (changes requested):
You're on Windows, right? Did you look at what needs to change in Makefile? The format and check entries need to change typeagent to src, and the mcp entry needs to use src/typeagent.

pyproject.toml
bmerkle: no, i removed it intentionally because it is the default setting, so we do not need to specify it.

bmerkle:
> You're on Windows, right? Did you look at what needs to change in Makefile? The format and check entries need to change typeagent to src, and the mcp entry needs to use src/typeagent.

yes. i should adopt the changes from make.bat to makefile, but it is a good hint thanks. so I will test on  Linux via my WSL.
 
I will also change makefile, but it contains a path i do not know ~/AISystems-Archive/data/knowpro/test/indexes

pyproject.toml
gvanrossum: What I saw earlier was a warning that the new version of the file didn't end in a newline. But now it's gone so I presume you managed to fix that. It doesn't look like the line changed otherwise.

gvanrossum (approved):
Perfect.

gvanrossum:
> Perfect.

Alas not. :-( I suspect you need to merge main first and then run tests locally to see what's failing.

bmerkle:
in get_keys.py i get a make format modification which was not commited in main branch.
it breaks it in several lines, not sure if CodeQL will accept that.

  ```
     self.credential = (
            DefaultAzureCredential()
        )  # CodeQL [SM05139] - This code is used as part of development setup only.

```

gvanrossum:
> in get_keys.py i get a make format modification which was not commited in main branch. it breaks it in several lines, not sure if CodeQL will accept that.
> 
> ```
>    self.credential = (
>           DefaultAzureCredential()
>       )  # CodeQL [SM05139] - This code is used as part of development setup only.
> ```

Rob is working on that in #131. For now ignore those errors.

bmerkle:
ok I have pushed. Should be fine now.

gvanrossum (approved):

gvanrossum:
I think your tests are failing because of #133. I will retest locally (I have a workaround) and then merge.

gvanrossum (approved):

gvanrossum:
Sorry, I can't merge this by myself. I need @robgruen's approval for my suggestions that were merged. (Re-approving doesn't help.)

robgruen (approved):

gvanrossum:
And it still won't merge. @robgruen can you merge this? Otherwise we're stuck.

bmerkle:
when we have merged this PR into main, I would create a new PR for the remaining topics of #125 

the PR is already here https://github.com/bmerkle/typeagent-py/tree/125 

I hope it is ok the make the remaining steps in one PR ?
- move test 125-tests 
- introduce samples 125-samples
- move spec-to-docs 125-spec

after that the structure should be ok, and we can continue with functional improvements

gvanrossum:
Can we just move test to tests? I'm not convinced we need a samples subdir (is that even a standard? and it feels like an odd word to me). The spec subdir probably needs to be deleted, not moved -- it was part of a failed AI experiment (the AI generated the spec but then wasn't able to implement it, and I couldn't get myself to edit it, it was way too verbose and repetitive). You can delete spec/ in a separate PR since it can be merged independent from everything else.

gvanrossum:
I wonder if we could make progress here if you did something like this:

- git checkout 125-src-create
- git diff main >125.diff
- git branch 125-redo
- git checkout 125-redo  # <----------------- IMPORTANT
- git apply 125.diff
- (git commit?)
- git push
- make a few PR from the new branch

That would have a cleaner history (a single commit, by you) and maybe if I approve it I can merge it.

gvanrossum:
Note I added a step after that comment sent an email -- use the version on GitHub, not the email (if you read those).

bmerkle:
yup, we can make 
- one PR with move of test to tests
- a second PR with rearrangement of samples, docs, specs, examples.  I described a bit in https://github.com/microsoft/typeagent-py/issues/125 what would make sense, but it do not know if there are any outdate files which should be deleted.
- IIRC most github repos have an "examples" in parallel to "src" where you can look how the framework is used etc. (so not "samples" but "examples"

gvanrossum:
Please make the one to move test -> tests (with tool updates including
Makefile).
For the other stuff there's not that much to gain from the rearrangement,
I'd like to wait on that. But go ahead and delete spec/ as a separate PR
(as it affects no tools).

bmerkle:
> I wonder if we could make progress here if you did something like this:
> 
> * git checkout 125-src-create
> * git diff main >125.diff
> * git branch 125-redo
> * git checkout 125-redo  # <----------------- IMPORTANT
> * git apply 125.diff
> * (git commit?)
> * git push
> * make a few PR from the new branch
> 
> That would have a cleaner history (a single commit, by you) and maybe if I approve it I can merge it.

i could squash it if that is want you mean ?

gvanrossum:
> i could squash it if that is want you mean ?

I've never used squash, but it sounds like it might work. Key requirement for a clean merge is (1) one diff, (2) authored by you only, and probably (3) in a new branch and PR.

bmerkle:
lets close/abandon this and use https://github.com/microsoft/typeagent-py/pull/139 instead.

PR #129, opened 2025-12-29
Make pyright error on unused variables and imports

gvanrossum:
- Fixed 23 unused variable errors.
  - This found a bug in answers.py: if N answers present, we'd add N copies of each of the generated answers, due to an unnecessary double loop. What was I thinking?
- Fixed 177 unused import errors.
- Fixed 1 duplicate import error.
- Renamed fixtures.py to conftest.py and remove most imports referencing it from the tests (fixtures defined in conftest.py are implicitly available in all tests).

I used AI (mostly Claude Opus 4.5) for this but reviewed everything it did and occasionally tweaked it (this is how I found the double loop in answers.py).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,4 +72,5 @@ testpaths = ["test"]
 
 [tool.pyright]
 reportUnusedVariable = true
+reportUnusedImport = true
 reportDuplicateImport = true
--- a/typeagent/emails/email_memory.py
+++ b/typeagent/emails/email_memory.py
@@ -5,11 +5,9 @@ import os
 from dataclasses import dataclass
 import json
 import copy
-from pydantic.dataclasses import dataclass as pydantic_dataclass
 import typechat
 from ..aitools import utils
 from ..knowpro import (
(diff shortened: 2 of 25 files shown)


gvanrossum:
The 'make format' failures are spurious -- I don't know what causes them (maybe black version skew?) and they don't happen on my Mac. The change looks harmless and not something that should trigger a reformat.

@robgruen this is ready for your review now.

gvanrossum:
> The 'make format' failures are spurious

No they weren't. There was a formatting issue in tools/get_keys.py due to the added warning -- black wants it formatted like this:

```py
        self.credential = (
            DefaultAzureCredential()
        )  # CodeQL [SM05139] - This code is used as part of development setup only.
```

I'm not sure if CodeQL is smart enough to honor that. Instead we should probably add `# fmt: skip` to the end of the line, or surround it with `# fmt: off` / `# fmt: on` .

@robgruen

robgruen:
I agree with you about CodeQL I'll make that change.

PR #131, opened 2025-12-29
Ignore formatting for CodeQL for security warning supression

robgruen:


robgruen:
> Did you check that CodeQl doesn't get upset by the extra comment tacked onto the end?

I've tried several variations, nothing seems to make the linux one happy. Windows one likes all of them.  Will need to investigate further, will let you know when it's ready.

gvanrossum:
We may have to upgrade black in pyproject.toml to "black>=25.12.0".

Alternatively, did you merge the *latest* main?

gvanrossum (approved):

Issue #133, opened 2025-12-29
Regression: when running `make test`, a dozen or more tests fail with "Connection failed."


gvanrossum:
This is new, likely introduced by #129, where some things changed around needs_auth and really_needs_auth?

gvanrossum:
It was introduced by [Remove unused imports in test/ and tools/](https://github.com/microsoft/typeagent-py/pull/129/changes/fd11e03bb182e6cfa2a2e8751313c44f80398f03). Now that's still a lot of changes. Which change was it?

gvanrossum:
It's interference between test_embeddings.py and test_podcasts.py. Presumably something changed in test_embeddings.py that broke auth.

gvanrossum:
Specifically, test/test_embeddings.py::test_refresh_auth does something that makes later auth attempts fail.

gvanrossum:
Fixed by #136.

PR #136, opened 2025-12-30
Fix Connection Errors caused by subtle change to test_embeddings.py

gvanrossum:
In test_embeddings.py, test_refresh_auth() does something to its embedding_model fixture that somehow hangs around past the end of the test, causing Connection Errors in later test files (not in its own file).

I've tried various ways to fix this; the cleanest way appears to be to put back the explicit import of the embedding_model fixture.

This fixes the problem when running tests locally, but in CI it still fails. I'm at my wits' end and plan to merge this despite the CI failures -- it's possible that the CI failures disappear once it's in main, and even if not, at least this fixes local test runs.

Tries to deal with #133.

### DIFF
--- a/test/test_embeddings.py
+++ b/test/test_embeddings.py
@@ -9,7 +9,8 @@ from pytest import MonkeyPatch
 import numpy as np
 
 from typeagent.aitools.embeddings import AsyncEmbeddingModel
-from conftest import embedding_model, fake_embeddings, fake_embeddings_tiktoken, FakeEmbeddings  # type: ignore
+from conftest import embedding_model  # type: ignore  # Magic, prevents side effects of mocking
+from conftest import FakeEmbeddings
 
 
 @pytest.mark.asyncio


gvanrossum:
Dang. In CI the tests fail in exactly the same way as before. Sigh.

robgruen:
I sent you a whole bunch of info in teams...can you check that out for my potential solution.

robgruen:
> I sent you a whole bunch of info in teams...can you check that out for my potential solution.

Ah, I see this is a different failure.  I haven't looked into this one yet.

gvanrossum:
That didn't help either.

gvanrossum:
> That didn't help either.

Oh, of course it didn't! The dist/ contains only things built from typegent/, and the only changes are in test/. So rolling back that ci.yml change.

I think I'm going to merge what's left -- it fixes the local tests at least, and who knows it may fix CI too once it's incorporated into main.

Issue #137, opened 2025-12-30
Test failures in CI even when there's a fix in a PR


gvanrossum:
We've had some mysteries now (#131, #136) where a PR doesn't appear to pick up the fixes to the test/ or tools/ directory until after it's been merged. (In one cases it was formatting a tool that failed; in another it was the test run.)

In both cases the scenario was as follows:

- There's an issue in a file under tools/ or test/.
- We develop a fix and test it locally; `make check test format` passes locally.
- However, in CI we get the same error that we got without the fix; pushing variations doesn't help.
- In desperation, we merge the PR even though CI fails.
- Magically, CI will now pass in following PRs, proving that the PR's fix indeed works!

I don't know why we're seeing this; possibly one of the recent security tweaks by @robgruen caused it. I *think* that it only occurs when no changes are made to the typeagent/ directory.

gvanrossum:
Fixed by #138. Contrary to my hypothesis above, we were not testing the PR branch at all! This fixes it.

PR #138, opened 2025-12-31
add ref when checking out

KRRT7:
for pull_request_target, we must explicitly checkout the PR's code without this, checkout defaults to the base branch (main), not the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
           exit 1  
 
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
       - name: Set up Python
         uses: actions/setup-python@v5


KRRT7:
closes  #137

gvanrossum:
Thanks! So does this mean that all this time we have been testing main instead of the submitted changes?

I am sure that I have seen buggy changes flagged by tests before. Maybe the typeagent package in dist/ is built from the branch?

I need to get off the beach to approve the tests.

gvanrossum:
It does explain the phenomena in #137.

gvanrossum (approved):

gvanrossum:
@KRRT7 Thank you so much for helping us out here!

KRRT7:
> Thanks! So does this mean that all this time we have been testing main instead of the submitted changes?

I'm not sure if you've always been doing that but what I think happened is that in #107 you added a custom pipeline which needs custom authentication, and to make that work securely you switched from pull_request to pull_request_target. The difference is that pull_request_target runs in the context of the base branch, not the PR — which is intentional for security (it prevents untrusted PR code from accessing secrets). However, the checkout GHA then defaults to checking out the base branch (main) unless you explicitly tell it to checkout the PR's code via the ref parameter. So since #107 merged, CI has been running against main while reporting results on the PR.

since you now require approval before CI run, it's safe to explicitly checkout the PR's code anyways.

gvanrossum:
Ah, I had thought #107 might have been the culprit. We've had about a dozen PRs since then. The tests now pass in CI, so it's really fixed. (I tested that errors introduced in a PR are now correctly causing CI to fail.)

IIUC the way things work then is that pull_request_target starts in main, does the security checks, and then ought to (and now finally does) check out the PR branch explicitly. Elegant design once you know how it works.

PR #139, opened 2025-12-31
add isort dependency and move typeagent to src

### LINKED ISSUE #125: migrate to python standard project layout

bmerkle:
@gvanrossum added isort dependency and moved typeagent to src 
this should fix #125 (first part with move to src)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
           exit 1  
 
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
       - name: Set up Python
         uses: actions/setup-python@v5


.github/workflows/ci.yml
```diff
@@ -78,13 +78,13 @@ jobs:
...
         run: |
-            uv run pyright typeagent test tools gmail
+            uv run pyright src test tools gmail
 
       - name: Run Format
         shell: bash
         if: matrix.task == 'format'
         run: |
-            uv run black -tpy312 -tpy313 -tpy314 typeagent test tools gmail demo --check
+            uv run black -tpy312 -tpy313 -tpy314 src test tools gmail demo --check
```
gvanrossum, line 89: Dang we need to add isort here too. I had forgotten CI isn't using 'make' any more (for a good reason: it doesn't work on Windows).

pyproject.toml
```diff
@@ -85,3 +85,8 @@ sections = ["FUTURE", "STDLIB", "THIRDPARTY", "AI", "FIRSTPARTY", "LOCALFOLDER"]
 known_ai = ["mcp", "openai", "pydantic", "pydantic_ai", "tiktoken", "typechat"]
 known_first_party = ["typeagent", "test", "tools", "gmail", "demo"]
 known_local_folder = ["conftest"]
+
+[dependency-groups]
+dev = [
+    "isort>=7.0.0",
+]
```
gvanrossum, line 92: Why do we need this? "isort>=7.0.0" is already listed above, under `[project.optional-dependencies]`. You automatically get it if you run `uv sync --extra dev` .

bmerkle: uv sync --extra Dev made this additional Line in this File

.github/workflows/ci.yml
bmerkle: OK will add it

pyproject.toml
gvanrossum: Interesting.

PR #140, opened 2025-12-31
TEST: Cause 'make check' to fail. See #138. DON'T MERGE

gvanrossum:


gvanrossum:
Yeah, this confirms it: we're using the main branch of at least tools (and I presume tests). Next I'll make a change to typeagent, see if that passes or fails.

gvanrossum:
Oh, type errors in typeagent are also not checked. So things really were quite bad.

gvanrossum:
Okay, after merging #138 things behave well. Closing this test PR.

Issue #134, opened 2025-12-29
Test failures send too much email


gvanrossum:
When tests fail, a separate email is sent for each of the 3x4 matrix items. This is rather verbose, especially since every time a 3rd party (without repo write access) the entire matrix fails. It's gotten worse with the latest security tightening.

@robgruen Could we have a lighter-weight approach? Or maybe test the matrix items in a nested loop in a single action rather than making 3x4 actions for them?

bmerkle:
there seems not to be a OOTB solution in ci.yml but we could go with the following approach:

1. Matrix jobs always succeed (or are allowed to fail)
2. Each job writes its result (pass/fail + metadata) as an artifact
3. A single follow-up job:
  - downloads artifacts
  - summarizes failures
  - fails once


```
jobs:
  test:
    strategy:
      fail-fast: false
      matrix:
        os: [ubuntu, windows, macos]
        python: ["3.10", "3.11", "3.12"]
    runs-on: ${{ matrix.os }}
    continue-on-error: true
    steps:
      - run: pytest || echo "failed" > result.txt
      - uses: actions/upload-artifact@v4
        with:
          name: result-${{ matrix.os }}-${{ matrix.python }}
          path: result.txt

  summarize:
    needs: test
    runs-on: ubuntu-latest
    steps:
      - uses: actions/download-artifact@v4
      - run: |
          failures=$(grep -R "failed" . || true)
          if [ -n "$failures" ]; then
            echo "$failures"
            exit 1
          fi`


```

gvanrossum:
I *think* we could cut the email barrage in half by not running CI again after a PR is merged. For this we could just remove these two lines from ci.yml:

```yaml
  push:
    branches: [ "main" ] 
```

@robgruen What do you think?

Not a full solution -- I am interested in @bmerkle's solution as well. But let's finish #125 first.

robgruen:
> I _think_ we could cut the email barrage in half by not running CI again after a PR is merged. For this we could just remove these two lines from ci.yml:
> 
>   push:
>     branches: [ "main" ] 
> [@robgruen](https://github.com/robgruen) What do you think?
> 
> Not a full solution -- I am interested in [@bmerkle](https://github.com/bmerkle)'s solution as well. But let's finish [#125](https://github.com/microsoft/typeagent-py/issues/125) first.

[...]

PR #141, opened 2025-12-31
Added gate and workflow_run trigger for CI

robgruen:
- Added gate workflow which checks user's permissions and returns success/failure based on their write access to the repo
- Updated CI workflow so that it is only triggered **after** gate workflow (can still be triggered manually)
- Left security checks in CI just in case it gets run without the Gate trigger.
(solution for #134 )


robgruen:
@gvanrossum-ms , we might need to force this one in since the **workflow_run** trigger won't work unless gate.yml is in the default branch.

gvanrossum:
In case you've been struggling with this, I just realized that the current workflow (at least on main) uses ci.yml from **main**. This follows from the pull_request_target semantics of starting with main.

robgruen:
Actually it's more a function of the workflow_run trigger.  By design it won't work unless the source of the target (in this case gate.yml) exists in the default branch.  This is by design for security.

gvanrossum:
Make this a draft PR?

robgruen:
> Make this a draft PR?

I think it's ready for approval.  I just merged main back in so I had to restart it but other than that it's ready to go.

PR #143, opened 2025-12-31
Better CI workflow with Permissions check as a short circuit

robgruen:


gvanrossum:
Now all we need is an outsider to push a test PR to verify that that doesn't run the tests. @bmerkle ?

bmerkle:
Will do the test with the move of Tests to new directory in next PR once i am Home in a few hours..... Happy new year to you all 😃🍀👋🎉

bmerkle:
ok PR #144 landed. @gvanrossum

gvanrossum:
The 3rd party experience is better, they get one failure of a job named permissions. (After an owner approves it.)

But as a project owner I still see 12 deployment approval emails?!

robgruen:
We could reduce this to 4 by moving the check + format steps to their own jobs within the CI.

Another option would be to update the "online" tests to only run for a specific Python version which would take it down to 1.  In the other cases we could do offline tests only.

Here's a draft PR for what the 1st option would look like: Moved Format and Check to their own jobs to reduce mail notifications for deployment approval by robgruen · Pull Request #148 · microsoft/typeagent-py<https://github.com/microsoft/typeagent-py/pull/148>

PR #145, opened 2026-01-01
125-examples: move various ancillary dirs  into subdirs

bmerkle:
- moved demo to examples/demo
- moved gmail to tools/gmail
- moved spec to docs/spec later we can delete/update outdated code/doc etc
- adjusted make and config files, ci, gitignore
- adjusted links in documentation
- removed unused import in simple_query_demo.py

fix part3+4 for #125

### DIFF
--- a/tests/test/test_transcripts.py
+++ b/tests/test/test_transcripts.py
@@ -67,7 +67,8 @@ def test_webvtt_timestamp_conversion():
 
 
 @pytest.mark.skipif(
-    not os.path.exists("tests/testdata/Confuse-A-Cat.vtt"), reason="Test VTT file not found"
+    not os.path.exists("tests/testdata/Confuse-A-Cat.vtt"),
+    reason="Test VTT file not found",
 )
 def test_get_transcript_info():
     """Test getting basic information from a VTT file."""
@@ -92,7 +93,8 @@ def conversation_settings(
 
 
 @pytest.mark.skipif(
-    not os.path.exists("tests/testdata/Confuse-A-Cat.vtt"), reason="Test VTT file not found"
+    not os.path.exists("tests/testdata/Confuse-A-Cat.vtt"),
+    reason="Test VTT file not found",
 )
 @pytest.mark.asyncio
 async def test_ingest_vtt_transcript(conversation_settings: ConversationSettings):


gvanrossum (approved):

bmerkle:
`uv run pyright src tests tools gmail`
it is failing because there is still gmail in that line (it should be "examples"

strange i do not have any line like this in my commit ?

gvanrossum:
> `uv run pyright src tests tools gmail`
> 
> it is failing because there is still gmail in that line (it should be "examples"
> 
> strange i do not have any line like this in my commit ?

This is because CI uses ci.yml from the main branch, always.

gvanrossum:
As I said, CI fails because it is using the old ci.yml from main. Please ignore the CI.

@bmerkle just fix the .gitignore and then I'll approve and merge.

I have an idea to make this a less frequent issue: we can put the list of folders to check and format in e.g. sources.txt file, and put in ci.yml something like `uv run pyright $(cat sources.txt)` and `uv run black <flags> $(cat sources)` -- the sources.txt file will come from the checked-out branch. But I'm not sure if Windows supports that syntax. :-( Anyway, if we were to do this we should do it later in a separate PR.

bmerkle:
> @bmerkle just fix the .gitignore and then I'll approve and merge.

have you made review for .gitignore ? i do not see a comment or requested changes in github UI
whats need to be fixed ?

.gitignore
```diff
@@ -30,13 +30,13 @@ pytest.local.ini
 
 # Evaluations
 /evals
-/testdata/Episode_53_Answer_results.json
-/testdata/Episode_53_Search_results.json
+/examples/testdata/Episode_53_Answer_results.json
+/examples/testdata/Episode_53_Search_results.json
```
gvanrossum: This should be /tests/ instead of /examples/

gvanrossum:
> > @bmerkle just fix the .gitignore and then I'll approve and merge.
> 
> have you made review for .gitignore ? i do not see a comment or requested changes in github UI whats need to be fixed ?

Sorry I did it on my phone which failed to send the comment. Redid it now. /examples/testdata/ should be /tests/testdata/.

gvanrossum:
Oh, looks like test_mcp.py is failing because it references a file in testdata, which was moved.

bmerkle:
> Oh, looks like test_mcp.py is failing because it references a file in testdata, which was moved.

right, should be fixed now. Currently the path are kind of hardcoded and we have some code duplication.
I think it makes sense to introduce a small utility method to have one place where we compute the test location files.

gvanrossum:
> I think it makes sense to introduce a small utility method to have one place where we compute the test location files.

Maybe a fixture that computes the path of testdata? Or even several for the paths of commonly used test data files?

Issue #146, opened 2026-01-01
Unify Testdata Path Computation


bmerkle:
We found out that Testdata Path Computation needs to be improved a bit:

>> I think it makes sense to introduce a small utility method to have one place where we compute the test location files.
> Maybe a fixture that computes the path of testdata? Or even several for the paths of commonly used test data files?

Currently we have three patterns:
1. Relative to test file (all test_*.py files)
2. Hardcoded relative to CWD (query.py and server.py... and also test_transripts.py)
3.  Relative to script dir (ingest_podcast.py)

I would refactor this, to fix the duplication and inconsistency. The implementation approach would use:
- pytest fixture (adding code and info to conftest.py)
- compute the testdata_path in only one place
- provide used test files as constants

My question is:
- how should we deal with non test scripts (like query.py or server.py)
- or better why does the production code like server.py include testdata at all, and wouldnt it be better to change the code of server.py to NOT contain any reference to test data

what do you think @gvanrossum ?

PR #147, opened 2026-01-01
Fixture testdatapath, refactor and unify testdatapath handling

bmerkle:
a possible solution for #146

for details/question of implementation approach, please check #146

### DIFF
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -94,6 +94,17 @@ async def test_mcp_server_query_conversation_slow(
     from mcp import ClientSession
     from mcp.client.stdio import stdio_client
 
+    # Pass through environment variables needed for authentication
+    # otherwise this test will fail in the CI on Windows only
+    if not (server_params.env) is None:
+        server_params.env.update(
+            {
+                k: v
+                for k, v in os.environ.items()
+                if k.startswith(("AZURE_", "OPENAI_")) or k in ("CREDENTIALS_JSON",)
+            }
+        )
+
     # Create client session and connect to server
     async with stdio_client(server_params) as (read, write):
         async with ClientSession(
@@ -126,7 +137,6 @@ async def test_mcp_server_query_conversation_slow(
             # Parse response (it should be JSON with success, answer, time_used)
             import json
 
-            print(f"Response text: {response_text}")
             try:
                 response_data = json.loads(response_text)
             except json.JSONDecodeError as e:


gvanrossum:
- I like what you did for conftest and the tests.
- I just realized that the tests are now in tests/test/test_foo.py. I had meant them to be in tests/test_foo.py (no "test/" folder in between). Sorry for not catching this in review.
- The shenanigans needed to import things from tests/test/conftest.py or from tools/util_testdata.py are horrible (and worse because isort doesn't understand '# fmt: off'), so let's not do that.
- I think maybe server.py and the tools should not have default paths built in at all. These will have to be passed in explicitly when we want to use them -- or they will have to continue to be hardcoded.

bmerkle:
> * I just realized that the tests are now in tests/test/test_foo.py. I had meant them to be in tests/test_foo.py (no "test/" folder in between). Sorry for not catching this in review.

oops, they should not be in extra subdir. Sorry, my fault. It meant to be /tests only. 
I will fix it. Seperate PR or in this PR ? i would prefer to fix that path issue in this PR directly

> * The shenanigans needed to import things from tests/test/conftest.py or from tools/util_testdata.py are horrible (and worse because isort doesn't understand '# fmt: off'), so let's not do that.

yes. I always feel bad when i do suppressions to hide hacks or workarounds like that. Will remove it. 

> * I think maybe server.py and the tools should not have default paths built in at all. These will have to be passed in explicitly when we want to use them -- or they will have to continue to be hardcoded.

passing explicit is a good idea, let me try that. (explicit is better than implicit ;-)

src/typeagent/mcp/server.py
```diff
@@ -161,24 +167,33 @@ async def make_context(
...
+        # Load from SQLite database
         conversation = await podcast.Podcast.create(settings)
+    elif podcast_index is not None:
+        # Load from JSON index files
+        conversation = await podcast.Podcast.read_from_file(podcast_index, settings)
+    else:
+        raise ValueError(
+            "Either --database or --podcast-index must be specified. "
+            "Use --podcast-index to specify the path to podcast index files "
+            f"(e.g., '{_EXAMPLE_PODCAST_INDEX}')."
```
gvanrossum, line 185: Oh, excellent. In practice the server almost always should use --database, which should be pre-filled by running one of the ingest tools.

src/typeagent/mcp/server.py
```diff
@@ -245,12 +260,49 @@ async def query_conversation(
...
-        help="Path to the SQLite database file (default: load from JSON file)",
+        help="Path to a SQLite database file with pre-indexed podcast data",
+    )
+    parser.add_argument(
+        "-p",
+        "--podcast-index",
+        type=str,
+        default=None,
+        help="Path to podcast index files (excluding '_data.json' suffix), "
+        f"e.g., '{_EXAMPLE_PODCAST_INDEX}'",
```
gvanrossum, line 271: At some point in the future we should be able to remove --podcast-index altogether, but not yet (and the test uses it).

tests/test_mcp_server.py
```diff
@@ -133,6 +126,7 @@ async def test_mcp_server_query_conversation_slow(
             # Parse response (it should be JSON with success, answer, time_used)
             import json
 
+            print(f"Response text: {response_text}")
```
gvanrossum: Do you still need this debug print?

tests/test/test_mcp_server.py
```diff
@@ -90,17 +94,6 @@ async def test_mcp_server_query_conversation_slow(
...
-    # otherwise this test will fail in the CI on Windows only
-    if not (server_params.env) is None:
-        server_params.env.update(
-            {
-                k: v
-                for k, v in os.environ.items()
-                if k.startswith(("AZURE_", "OPENAI_")) or k in ("CREDENTIALS_JSON",)
-            }
-        )
-
```
gvanrossum, line 103: I hope that you're right that we don't need this any more.

gvanrossum (approved):

tests/test/test_mcp_server.py
bmerkle: Ups, sorry. This is still necessary for ci in windows. I will put it pack.

tests/test_mcp_server.py
bmerkle: removed

gvanrossum (approved):

PR #148, opened 2026-01-02
Moved Format and Check to their own jobs to reduce mail notifications for deployment approval

robgruen:
Less jobs awaiting approval with this layout:

image


.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ name: CI
 
 on:
   push:
-    branches: [ "main" ] 
+    branches: [ "main", "dev/robgruen/workflow_update3" ]
```
gvanrossum: @robgruen Can you delete your branch from the list? Then I can merge it for you. (If I were to make the change myself we might end up in a situation where only a 3rd owner can merge it...)

gvanrossum: Maybe also un-draft it. But only when you're happy with this step. (I am!)

robgruen: done.

gvanrossum:
Closing in favor of #151.

PR #150, opened 2026-01-03
Fix #149 pydantic configuration / setup to cover remaining warnings/errors 

### LINKED ISSUE #149: Fix pydantic configuration / setup to cover remaining warnings/errors
We have quite a few pydantic warnings/error which should be fixed:

e.g. 
```
@dataclass(unsafe_hash=True)
class Term:
    """A # Term must be hashable to allow using it as a dict key or set member."""

    text: str
    weight: float | None = None  # Optional weighting for these matches.

```

at all call sites of Term i get e.g.
```
def test_termset_add():
    """Test adding terms to the TermSet."""
    term1 = Term(text="example1", weight=1.0)

```  
No parameter named "text"PyrightreportCallIssue
No parameter named "weight"PyrightreportCallIssue

I will fix pydantic configuration/setup to get the right analysis setup. The code (in this case) is fine, but the analysis is not setup correct.

We might also have other places where **real** type errors could happen but we currently do not see them because of the other cases/warnings.

bmerkle:
- fixed No parameter named "xyz "PyrightreportCallIssue for dataclasses, e.g. class Term
- fix for #149

### DIFF
--- a/src/typeagent/knowpro/interfaces_core.py
+++ b/src/typeagent/knowpro/interfaces_core.py
@@ -4,7 +4,6 @@
 
 from __future__ import annotations
 
-from dataclasses import dataclass as stdlib_dataclass
 from datetime import datetime as Datetime
 from typing import (
     Any,
@@ -17,13 +16,7 @@ from typing import (
     TypedDict,
 )
 
-if TYPE_CHECKING:
-    # Use stdlib dataclass for type checking (pyright understands this)
-    dataclass = stdlib_dataclass
-else:
-    # Use pydantic dataclass at runtime for serialization support
-    from pydantic.dataclasses import dataclass
-
+from pydantic.dataclasses import dataclass
 import typechat
 
 from . import kplib


src/typeagent/knowpro/interfaces_core.py
```diff
@@ -16,7 +17,13 @@
     TypedDict,
 )
 
-from pydantic.dataclasses import dataclass
+if TYPE_CHECKING:
+    # Use stdlib dataclass for type checking (pyright understands this)
+    dataclass = stdlib_dataclass
```
gvanrossum: Why not `from dataclasses import dataclass`? Does that not work?

bmerkle: the real issue was the VSCode plugin and the pyright version shipped with it.
I reverted this change and added instead information how to setup VSCode in the right way

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # Python
 __pycache__
+.venv/
```
gvanrossum, line 6: On Linux, the .venv/ folder already contains its own .gitignore containing the single line `*`. Does it not have that on Windows?

gvanrossum (approved):
I do wonder if things wouldn't be much easier if we just recommended not to install the standalone Pyright plugin and rely on Pylance? That's what I do and I don't see the problem. I have not messed with the Pylance version. But it's quite possible that on Windows things are different. (I can't test that right now.)

I'll just merge and we can tweak later.

.gitignore
bmerkle: AFAIK "uv init" only generates the .gitignore in the .venv folder. 
it is somehow not regenerated if i throw away the .venv and then "uv sync" it again (this is what i did)

so most projects .gitignore end up with a .venv/ line in the end unfortunately. (a bit of redundance), because only the initial generator or project setup step created the file but it is not recreated in later lifecycle steps.
so the end it is not a windows or linux problem but a uv, poetry, pixi setup feature ;-)

gvanrossum: > AFAIK "uv init" only generates the .gitignore in the .venv folder. it is somehow not regenerated if i throw away the .venv and then "uv sync" it again (this is what i did)

Interesting. Seems to be another difference between .venv/ on Unix and Windows. Thanks for speaking up for Windows!

PR #151, opened 2026-01-04
CI with Online/Offline tests

robgruen:
image


robgruen:
@gvanrossum-ms , I'm assuming the tests will just skip if there are no keys (I was lazy and didn't check).  If not, we'll want to make that update before merging this.

gvanrossum:
> @gvanrossum-ms , I'm assuming the tests will just skip if there are no keys (I was lazy and didn't check).

Yes, any test that needs a secret uses the `really_needs_auth` fixture which skips the test if no secret is found in the expected environment variables.

gvanrossum (approved):
Looks good. Let's try this!

PR #162, opened 2026-01-14
Various ergonomic improvements to email ingestion

### LINKED ISSUE #166: SQLite create_storage_provider() is slow when database has data already
What is it doing? I have a db with ~2500 messages and create_storage_provider() took 230 seconds! Something must be quadratic or doing too many small db queries or going out to an embedding server?

gvanrossum-ms:
* Generalize IngestedSources to store a status field for each source_id.

  This changes the schema for IngestedSources, adding a text column 'status' that can describe whether indexing of a source_id succeeded, failed, or resulted in some other status. There is one well-known value, 'ingested' (STATUS_INGESTED), but using `mark_source_ingested()` it can be set to any other string value, and the status can be retrieved using `get_source_status()`. The existing `is_source_ingested()` returns True only if the status field has the exact value STATUS_INGESTED.

  To alter a precious existing database, do this using sqlite3:
  ```
  ALTER TABLE IngestedSources
  ADD COLUMN status TEXT NOT NULL DEFAULT 'ingested';
  ```
  That sets the status for every row present to 'ingested', matching the original behavior. (The in-memory storage provider implements the API but always reports the status to be 'ingested' when set.)

* Increase TypeChat timeout_seconds from 10 (default) to 30 (DEFAULT_TIMEOUT_SECONDS).

* Switch from email_id to filename as source_id -- this allows us to skip parsing the message when it's already ingested.

* Remove quadratic behavior from SqliteRelatedTermsFuzzy and SqliteMessageTextIndex  constructors.

* Change signature of VectorBase.add_embeddings() -- a new argument keys precedes the embeddings argument. (Less backwards compatible, but consistent with add_embedding().)

* Change output of utils.timelog(): label shows first and immediately, time follows; and all goes to stderr.

* Streamlining and improvements to ingest_email.py verbose output, e.g. print only one line for already-ingested files, and decode RFC 2047 encoded-word strings.

* Fixes #166 (quadratic behavior in SqliteRelatedTermsFuzzy and SqliteMessageTextIndex).

### DIFF
--- a/tools/load_json.py
+++ b/tools/load_json.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""
+JSON to SQLite Database Loader
+
+This tool loads a JSON-serialized podcast database into a SQLite database
+that can be queried using tools/query.py.
+
+Usage:
+    python tools/load_json.py <index_path> --database <db_file>
+    python tools/load_json.py tests/testdata/Episode_53_AdrianTchaikovsky_index -d transcript.db
+
+The index_path should exclude the "_data.json" suffix.
+"""
+
+import argparse
+import asyncio
+import os
+
+from typeagent.aitools import utils
+from typeagent.knowpro.convsettings import ConversationSettings
+from typeagent.podcasts import podcast
+from typeagent.storage.utils import create_storage_provider
+
+
+async def load_json_to_database(
+    podcast_file_prefix: str,
+    dbname: str,
+    verbose: bool = False,
+) -> None:
+    """Load JSON-serialized podcast data into a SQLite database.
+
+    Args:
+        podcast_file_prefix: Path to podcast index files (without "_data.json" suffix)
+        dbname: Path to SQLite database file (must be empty)
+        verbose: Whether to show verbose output
+    """
+    if verbose:
+        print(f"Loading podcast from JSON: {podcast_file_prefix}")
+        print(f"Target database: {dbname}")
+
+    # Create settings and storage provider
+    settings = ConversationSettings()
+    settings.storage_provider = await create_storage_provider(
+        settings.message_text_index_settings,
+        settings.related_term_index_settings,
+        dbname,
+        podcast.PodcastMessage,
+    )
+
+    # Get the storage provider to check if database is empty
+    provider = await settings.get_storage_provider()
+    msgs = await provider.get_message_collection()
+
+    # Check if database already has data
+    msg_count = await msgs.size()
+    if msg_count > 0:
+        raise RuntimeError(
+            f"Database '{dbname}' already contains {msg_count} messages. "
+            "The database must be empty to load new data. "
+            "Please use a different database file or remove the existing one."
+        )
+
+    # Load podcast from JSON files
+    with utils.timelog(f"Loading podcast from {podcast_file_prefix!r}"):
+        async with provider:
+            conversation = await podcast.Podcast.read_from_file(
+                podcast_file_prefix, settings, dbname
+            )
+
+    # Print statistics
+    if verbose:
+        print(f"\nSuccessfully loaded podcast data:")
+        print(f"  {await conversation.messages.size()} messages")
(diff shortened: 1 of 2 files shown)


robgruen:
Why have a string column instead of a new table with status and reference the id?  Probably more complicated if you added the functionality for tracking new status ids for custom status fields, but should be less storage consumed in the long run.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -141,13 +141,21 @@ async def add_messages_with_indexing(
...
         storage = await self.settings.get_storage_provider()
+        if source_ids:
+            if len(source_ids) != len(messages):
+                raise ValueError(
+                    f"Length of source_ids {len(source_ids)} "
+                    f"must match length of messages {len(messages)}"
+                )
+            async with storage:
+                for source_id in source_ids:
+                    storage.mark_source_ingested(source_id, "failed")
```
robgruen: Are we forcing a specific version of python?  

Can we use string enums if so?

src/typeagent/storage/sqlite/schema.py
```diff
@@ -143,7 +143,8 @@
 # This prevents re-ingesting the same content on subsequent runs
 INGESTED_SOURCES_SCHEMA = """
 CREATE TABLE IF NOT EXISTS IngestedSources (
-    source_id TEXT PRIMARY KEY      -- External source identifier (email ID, file path, etc.)
+    source_id TEXT PRIMARY KEY,      -- External source identifier (email ID, file path, etc.)
+    status TEXT NOT NULL DEFAULT 'ingested'  -- Status of the source (e.g., 'ingested', 'failed')
```
robgruen: See my previous comment about making this a number/id to a status column with the status description.

tools/ingest_email.py
```diff
@@ -83,6 +91,20 @@ def collect_email_files(paths: list[str], verbose: bool) -> list[Path]:
...
 
+def decode_encoded_word(s: str) -> str:
+    """Decode an RFC 2047 encoded string."""
+    if "=?utf-8?" not in s:
+        return s  # Fast path for common case
+    decoded_parts = decode_header(s)
+    decoded_string = ""
+    for part, encoding in decoded_parts:
+        if isinstance(part, bytes):
+            decoded_string += part.decode(encoding or "utf-8", errors="replace")
```
robgruen: nit: .join instead of concat.

robgruen (approved):

tools/ingest_email.py
gvanrossum: Oh, booh Copilot. Will fix.

src/typeagent/knowpro/conversation_base.py
gvanrossum: Enums are stable enough in the versions we support (3.12+). But I've changed my mind slightly. In add_messages_with_indexing() we should only mark positive results. This is more backwards compatible.

Then in ingest_email.py, in the `except Exception` clause, we should mark the file with the full name of the exception that was caught. This results in fewer transactions (one if success; two if failure) and makes it easy to (manually) delete entries due to a certain exception. (E.g., occasionally I see Azure exceptions complaining that the token is invalid -- when that happen, all remaining files are marked as failed in rapid succession.)

I don't see a role for an enum, however, I thing we can have a CONSTANT for the string "ingested".

gvanrossum:
> Why have a string column instead of a new table with status and reference the id? Probably more complicated if you added the functionality for tracking new status ids for custom status fields, but should be less storage consumed in the long run.

I somehow doubt that this table is going to be one of the bigger ones -- when I get back to the machine where I am ingesting thousands of emails I will measure all tables. (Based on a small sample, I suspect that tables with embeddings like RelatedTermsFuzzy will be by far the largest.)

We could use NULL to represent "ingested", that would probably use up less space than a string, but I'm reluctant to play too many games here unless it's a known bottleneck. What I really do want is the semantics of storing various status strings for failures (see other comment).

src/typeagent/knowpro/conversation_base.py
gvanrossum: Making it a constant.

Issue #121, opened 2025-12-23
tools/release.py should update typeagent version in pyproject.toml and uv.lock


gvanrossum:
This currently falls on the first commit *after* the release.

Problem: Currently the release script doesn't create any commits, so all we have to do is push tags. Pushing commits to main is problematic due to the way access control is set up.

gvanrossum:
Possibly it could create a new "release-x.y.z" branch and make the commit as well as the tag there. It could even make the PR. I would then have to have a coworker approve that before I can merge it in main.

gvanrossum-ms:
It should probably also bump the project version in pyproject.toml to something like 0.3.4+.

iamrajiv:
hi @gvanrossum can you review this https://github.com/microsoft/typeagent-py/pull/169? so when you run the release script it will now commit both pyproject.toml and uv.lock together

gvanrossum:
Reviewed. But you're not doing everything I suggested here -- I'd like a separate commit, *after* the tag is created, that adds a '+' to the version in pyproject.toml, so it's clear that versions built from main are not *identical* to the latest release. I also suggested it create a new branch + PR (that would have to be approved and merged manually). So let's not close this issue just yet.

iamrajiv:
Updated the PR to address the additional feedback. The release script now:

1. Creates a `release-X.Y.Z` branch instead of committing directly to main
2. Bumps version to X.Y.Z and updates uv.lock
3. Creates the tag
4. Bumps version to X.Y.Z+ for development
5. Pushes the branch and creates a PR via `gh` CLI

@gvanrossum ready for another look when you get a chance

gvanrossum:
Thanks again!

PR #169, opened 2026-01-20
chore: Enhance release script to update uv.lock and create a release branch+PR

iamrajiv:
Fixes: #121

### DIFF
--- a/tools/release.py
+++ b/tools/release.py
@@ -10,7 +10,7 @@ This script:
 2. Bumps the patch version (3rd part) in pyproject.toml, or sets the whole version
 3. Updates uv.lock to reflect the version change
 4. Commits the change and creates a git tag in the format "v{major}.{minor}.{patch}-py"
-5. Bumps version to X.Y.Z+ (post-release development marker)
+5. Bumps version to X.Y.Z.dev (post-release development marker)
 6. Updates uv.lock again for the post-release version
 7. Commits the post-release version change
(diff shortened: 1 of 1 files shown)


tools/release.py
```diff
@@ -239,6 +260,12 @@ def main():
             )
             return 1
 
+    # Check that uv is available
+    if not check_uv_available():
+        print("Error: 'uv' command not found. Please install uv first.", file=sys.stderr)
+        print("  Install with: curl -LsSf https://astral.sh/uv/install.sh | sh")
```
gvanrossum: Why doesn't this go to stderr too?

gvanrossum (commented):
Thanks again! LG, just one nit.

tools/release.py
```diff
@@ -320,20 +379,68 @@ def main():
         print(f"Error: Failed to create tag {tag_name}", file=sys.stderr)
         return 1
 
-    # Push tags
+    # Post-release: bump version to X.Y.Z+
+    post_release_version = f"{new_version}+"
```
gvanrossum: It turns out you can't have a + sign in a version number -- uv balks when reading that back from pyproject.toml. However, you *can* add ".dev". However, then this tool can't parse the version! So you need to make two changes:
- replace "+" with ".dev" here
- fix the parsing so it strips ".dev" from the end of the version when it sees it.

PR #205, opened 2026-02-26
Remove "inane" docstrings, by Opus 4.6

gvanrossum-ms:
This took **12 minutes** of Opus 4.6 time and the result is that it deleted selected one-line docstrings.

I'm not sure if I like this. @bmerkle? In any case I propose to keep it out of the 0.4.0 release.


gvanrossum:
No longer interested.

PR #229, opened 2026-04-10
perf: Defer black import to first use

KRRT7:
**Stack: 2/4** — depends on #231. Merge #231 first, then this PR.

---

- Defer `import black` from module level to first use in `answers.py` and `utils.py`
- `black` (code formatter + transitive deps: pathspec, black.nodes, etc.) loaded on every `import typeagent` but only used in two cold formatting paths

`black.format_str()` is called in two places:
- `create_context_prompt()` in `knowpro/answers.py` — formats debug context for LLM prompts
- `format_code()` in `aitools/utils.py` — developer pretty-print utility

Neither runs during normal library operation. Moving the import inside each function eliminates ~78ms of transitive module loading from the import chain.

|:---|---:|---:|---:|
| `import typeagent` | 791 ms ± 11 ms | 713 ms ± 8 ms | **1.11x** |

#### Offline E2E Test Suite (hyperfine, warmup 2, min-runs 10)

|:---|---:|---:|---:|
| 69 offline tests | 5.72 s ± 90 ms | 5.60 s ± 98 ms | 1.02x |

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["uv_build>=0.9.10,<0.10.0"]
+requires = ["uv_build>=0.9.10,<0.11.0"]
 build-backend = "uv_build"
 
 [project]
--- a/src/typeagent/aitools/utils.py
+++ b/src/typeagent/aitools/utils.py
@@ -11,7 +11,6 @@ import shutil
 import sys
 import time
 
-import black
 import colorama
 
 import typechat
@@ -57,6 +56,8 @@ def format_code(text: str, line_width=None) -> str:
 
     NOTE: The text must be a valid Python expression or code block.
     """
+    import black
+
     if line_width is None:
         # Use the terminal width, but cap it to 200 characters.
         line_width = min(200, shutil.get_terminal_size().columns)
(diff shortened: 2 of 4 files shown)


gvanrossum (approved):
Should have one this a long time ago. :-(

KRRT7:
Thanks, wasn't sure you wanted import time optimizations, if you do, I have a few more ready to go.

gvanrossum:
This one is dear to my heart because I wish we could remove black from the mandatory dependencies. Happy to review more.

KRRT7:
why can't you remove black? out of curiousness

gvanrossum:
> why can't you remove black? out of curiousness

I could, but it gives the nicest formatted expressions when doing hardcore debugging. What we could do is import it conditionally and just use repr() if black cannot be imported. Then we can remove it from the main deps. Developers have the dev tools installed (pyright, black, isort, pytest, etc.) and will see the black-formatted debug output.

KRRT7:
Went ahead and did this — #235 replaces `black.format_str` with `pprint.pformat` + `ast.literal_eval` in both call sites (stdlib only, no conditional imports). Moved `black` from `dependencies` to the `dev` dependency-group so `make format`/`make check` still work but library consumers don't pull it in.

PR #230, opened 2026-04-10
perf: Batch SQLite INSERTs for indexing pipeline

KRRT7:
**Stack: 3/4** — depends on #229. Merge #231, #229, then this PR.

---

- Add `add_terms_batch` and `add_properties_batch` to `ITermToSemanticRefIndex` and `IPropertyToSemanticRefIndex` interfaces
- SQLite backend uses `executemany` instead of individual `cursor.execute()` calls (~1000+ calls per indexing batch reduced to 2-3)
- Restructure `add_metadata_to_index_from_list` and `add_to_property_index` to collect all data first (pure functions), then batch-insert
- Memory backend implements batch methods as loops for interface compatibility

#### Indexing Pipeline (pytest-async-benchmark pedantic, 20 rounds, 3 warmup)

Only the hot path (`add_messages_with_indexing`) is timed -- DB creation, storage init, and teardown are excluded.

| Benchmark | Before (min) | After (min) | Speedup |
|:---|---:|---:|---:|
| `add_messages_with_indexing` (200 msgs) | 28.8 ms | 25.0 ms | **1.16x** |
| `add_messages_with_indexing` (50 msgs) | 7.8 ms | 6.7 ms | **1.16x** |
| VTT ingest (40 msgs) | 6.9 ms | 6.1 ms | **1.14x** |

Consistent ~14-16% improvement -- `executemany` amortizes per-call overhead.

<details>
<summary><b>Reproduce the benchmark locally</b></summary>

Save the benchmark file below as `tests/benchmarks/test_benchmark_indexing.py`, then:

```bash
pip install 'pytest-async-benchmark @ git+https://github.com/KRRT7/pytest-async-benchmark.git@feat/pedantic-mode' pytest-asyncio

# Run on main
git checkout main
python -m pytest tests/benchmarks/test_benchmark_indexing.py -v -s

# Run on this branch
git checkout perf/batch-inserts
python -m pytest tests/benchmarks/test_benchmark_indexing.py -v -s
```
</details>

---


gvanrossum (commented):
Thanks for this! It requires more review time than I have right now, so I'll keep it open until I have more time.

src/typeagent/storage/memory/semrefindex.py
```diff
@@ -585,18 +629,50 @@ async def add_metadata_to_index_from_list[TMessage: IMessage](
     knowledge_validator: KnowledgeValidator | None = None,
 ) -> None:
     """Extract metadata knowledge from a list of messages starting at ordinal."""
+    next_ordinal = await semantic_refs.size()
```
bmerkle, line 541: I think there is a possible bug because the new implementation add_metadata_to_index_from_list drops inverse_actions

The original per-item functions (add_knowledge_to_index, add_knowledge_to_index) process knowledge_response.inverse_actions in addition to actions. 

The new batched add_metadata_to_index_from_list (semrefindex.py:631-675) only iterates over entities, actions, and topics — inverse_actions are silently skipped. This is a correctness regression: any inverse actions in messages will no longer be indexed.

Maybe we also should extend the testcases to cover all the knowlege_schemas.

src/typeagent/storage/memory/propindex.py
```diff
@@ -109,6 +110,63 @@ async def build_property_index(conversation: IConversation) -> None:
...
+    ]
+    for t in entity.type:
+        props.append((PropertyNames.EntityType.value, t, ordinal))
+    if entity.facets:
+        for facet in entity.facets:
+            props.extend(collect_facet_properties(facet, ordinal))
+    return props
+
+
+def collect_action_properties(
```
bmerkle, line 147: potential BUG: collect_action_properties drops subject_entity_facet

The old add_action_properties_to_index does not index action.subject_entity_facet (correct — only the semrefindex adds facet terms). But the new collect_action_properties in propindex.py also doesn't — so this is fine. 
However, comparing the semrefindex side: collect_action_terms does include action.subject_entity_facet via collect_facet_terms, which matches the old add_action_to_index.

src/typeagent/storage/memory/semrefindex.py
bmerkle, line 449: add_metadata_to_index is not batched

Only add_metadata_to_index_from_list was converted to batch mode. The older add_metadata_to_index (which takes AsyncIterable[TMessage], semrefindex.py:547-580) still uses single-item inserts. This is inconsistent — callers using the iterator-based path won't benefit from the optimization. This may be intentional (harder to batch an async iterator), but it's a missed optimization.

src/typeagent/storage/sqlite/propindex.py
```diff
@@ -67,6 +68,43 @@ async def add_property(
...
+            tuple[
+                str,
+                str,
+                interfaces.SemanticRefOrdinal | interfaces.ScoredSemanticRefOrdinal,
+            ]
+        ],
+    ) -> None:
+        if not properties:
+            return
+        from ...storage.memory.propindex import (
```
bmerkle: Don't put imports inside functions
these should be top-level imports, consistent with the coding guidelines

see AGENTS.md

src/typeagent/storage/memory/semrefindex.py
```diff
@@ -577,6 +579,48 @@ async def add_metadata_to_index[TMessage: IMessage](
...
+                terms.append(param)
+            else:
+                terms.append(param.name)
+                if isinstance(param.value, str):
+                    terms.append(param.value)
+    terms.extend(collect_facet_terms(action.subject_entity_facet))
+    return terms
+
+
 async def add_metadata_to_index_from_list[TMessage: IMessage](
```
bmerkle, line 533: add_metadata_to_index_from_list also doesn't process inverse_actions from knowledge_response

src/typeagent/storage/memory/semrefindex.py
bmerkle: We have massive code duplication.  This is not origin in this PR but IMO we should fix this.

There are two parallel sets of functions that do the same thing with minor variations:

add_entity_to_index (line 154) vs add_entity (line 199)
add_action_to_index (line 486) vs add_action (line 327)
add_topic_to_index (line 468) vs add_topic (line 277)
add_facet at line 243 is shared but called differently
add_knowledge_to_index (line 520) vs add_knowledge_to_semantic_ref_index (line 420)

The _to_index variants use text_range_from_location while the other set uses text_range_from_message_chunk and supports terms_added. 

Now the PR adds a third path (collect_*_terms + batch). Every behavioral change must be synchronized across all three

bmerkle: please see comment on line R154

bmerkle: please see comment on line R154

bmerkle, line 206: please see comment on line R154

bmerkle, line 429: please see comment on line R154

bmerkle (commented):
Hi @KRRT7 
I was asked by @gvanrossum to do a review of this PR
Please find attached some comments. 

There are also some pre-existing issues in these files, so things which have not be introduced by this PR, but i would suggest that we cover those in a future, seperate PR.
I have only mentioned the code duplicate, which could possibly be fixed also in this PR.

please let me know what you think.

src/typeagent/storage/sqlite/propindex.py
KRRT7: Hi, I think that there are times when imports *should* go inside functions, one such example is https://github.com/microsoft/typeagent-py/pull/229/changes
however, I'll need to double check if it really matters for this PR.

- as an aside I wanted to take a look to your agentic rules in a separate PR to see how I could optimize them as well for your needs.

KRRT7:
> Hi @KRRT7
I was asked by @gvanrossum to do a review of this PR

Thanks! I'll take a closer look to your reviews this afternoon.

bmerkle:
Hi @KRRT7 can you please review the open comments from me ? Thanks :-)

bmerkle (commented):
@KRRT7 please check my comments

src/typeagent/storage/memory/semrefindex.py
KRRT7: Fixed. `inverse_actions` are now indexed in `add_metadata_to_index_from_list` — added a dedicated loop matching the same pattern as `actions`. Also fixed the same bug in `add_metadata_to_index` (the async iterator path) which had the same omission pre-existing.

Additionally unified the duplicated functions: removed `add_entity_to_index`, `add_action_to_index`, `add_topic_to_index`, and `text_range_from_location` — they were identical to `add_entity`, `add_action`, `add_topic`, and `text_range_from_message_chunk` respectively. Updated all callers (tests included). This eliminates the risk of behavioral drift between parallel paths.

src/typeagent/storage/memory/propindex.py
KRRT7: Confirmed — not a bug. `collect_action_properties` correctly omits `subject_entity_facet` (only the semrefindex adds facet terms, not the property index). The semrefindex side (`collect_action_terms`) does include it via `collect_facet_terms`, matching the original `add_action_to_index` behavior.

src/typeagent/storage/memory/semrefindex.py
KRRT7: Intentional — batching an `AsyncIterable` requires buffering the full stream which defeats the purpose of iterators. The iterator path (`add_metadata_to_index`) is used for the initial full-conversation build where messages stream in; the list path (`add_metadata_to_index_from_list`) is the hot incremental path used by `add_messages_with_indexing` and benefits most from batching.

That said, I did fix `add_metadata_to_index` to also process `inverse_actions` — it had the same omission as the list path.

src/typeagent/storage/sqlite/propindex.py
KRRT7: Fixed. Moved `make_property_term_text` and `split_property_term_text` imports to top-level in `sqlite/propindex.py`, consistent with AGENTS.md guidelines. All three call sites (`add_property`, `add_properties_batch`, `lookup_property`) now use the top-level import.

src/typeagent/storage/memory/semrefindex.py
KRRT7: Fixed — see reply on the main inverse_actions thread. Both `add_metadata_to_index_from_list` and `add_metadata_to_index` now process `inverse_actions`.

src/typeagent/storage/memory/semrefindex.py
KRRT7: Addressed in this update. The three parallel paths have been reduced to two:

1. **`add_entity` / `add_action` / `add_topic`** — the single authoritative implementation, used by the per-item paths (`add_knowledge_to_semantic_ref_index`, `add_knowledge_to_index`, `add_metadata_to_index`)
2. **`collect_*_terms` + batch** — the batched path for `add_metadata_to_index_from_list` (needed for `executemany` performance)

Deleted: `add_entity_to_index`, `add_action_to_index`, `add_topic_to_index`, `text_range_from_location` (were duplicates of path 1). `add_knowledge_to_index` now delegates to path 1. All callers updated.

The `collect_*_terms` functions (path 2) remain because the batch path needs to pre-collect terms before a single `add_terms_batch` call — this is the whole point of the perf optimization. But they only need to stay in sync with the `add_*` functions, not maintain a third independent implementation.

KRRT7:
@bmerkle Thanks for the thorough review — all comments addressed in `9ae667f`. Summary:

- **Fixed `inverse_actions` bug** — both `add_metadata_to_index_from_list` (regression) and `add_metadata_to_index` (pre-existing) now index inverse actions
- **Eliminated code duplication** — deleted `add_entity_to_index`, `add_action_to_index`, `add_topic_to_index`, and `text_range_from_location` (were duplicates of `add_entity`/`add_action`/`add_topic`/`text_range_from_message_chunk`). Reduced from 3 parallel paths to 2. Updated all callers and tests.
- **Moved imports to top-level** in `sqlite/propindex.py` per AGENTS.md

`make check test format` passes (479/479 offline tests green, 3 failures are API rate-limits).

Ready for re-review.

bmerkle (approved):
@KRRT7 sorry for the delay.... LGTM, thanks for the cleanup and the fixes.  
all reviews comments have been covered and changes were implemented.

PR #231, opened 2026-04-10
Fix parse_azure_endpoint passing query string to AsyncAzureOpenAI

KRRT7:
**Stack: 1/4** — merge this first, then #229, #230, #232.

---

- `parse_azure_endpoint` returned the full URL including `?api-version=...`
- `AsyncAzureOpenAI` appends `/openai/` to `azure_endpoint`, producing a mangled URL with the query string in the path
- Now strips the query string with `str.split("?", 1)[0]` before returning
- Added 6 unit tests covering: basic URL, no version, separate env var, missing env var, empty query string

## Benchmark

No performance impact — this is a correctness fix.

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["uv_build>=0.9.10,<0.10.0"]
+requires = ["uv_build>=0.9.10,<0.11.0"]
 build-backend = "uv_build"
 
 [project]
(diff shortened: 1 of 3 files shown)


gvanrossum (approved):
Looks great!

bmerkle:
this break current setup, e.g. .env files containing URLs with api version information (which is the default in Azure Foundry setup). 
please see bugreport #238

PR #233, opened 2026-04-10
Add color and improve rendering in chat_sessions.py tool

gvanrossum:
@bmerkle could you review this too? This is really an independent little tool, but super handy -- it can show you all your AI chats used from VS Code. (I suspect it wouldn't be too hard to be able to point it to other frameworks' logs as well.)

### DIFF
--- a/tools/chat_sessions.py
+++ b/tools/chat_sessions.py
@@ -31,9 +31,32 @@ import termios
 import textwrap
 from typing import Any
 
+from colorama import Fore, init, Style
+
 VSCODE_USER_DIR = Path.home() / "Library" / "Application Support" / "Code" / "User"
 # Linux: Path.home() / ".config" / "Code" / "User"
 # Windows: Path.home() / "AppData" / "Roaming" / "Code" / "User"
+
+# Color settings
+use_color = True
+
+
+def should_use_color(args: argparse.Namespace | None = None) -> bool:
+    """Determine if color should be used based on args and environment."""
+    # Check explicit command-line flags first
+    if args is not None:
+        if hasattr(args, "color"):
+            if args.color == "always":
+                return True
+            if args.color == "never":
+                return False
+    # Check environment variables
+    if os.environ.get("NO_COLOR"):
+        return False
+    if os.environ.get("FORCE_COLOR"):
+        return True
+    # Default: use color if output is a TTY
+    return sys.stdout.isatty()
 
 
 def find_session_dirs() -> list[Path]:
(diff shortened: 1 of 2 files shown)


tools/chat_sessions.py
```diff
@@ -18,20 +18,98 @@
 from collections.abc import Iterator
 import contextlib
 import datetime
+import fcntl
```
bmerkle: I am also testing on Windows and currently chat_session.py is Unix only unfortunately.

import fcntl, import termios are Unix-only modules

tools/chat_sessions.py
```diff
@@ -18,20 +18,98 @@
...
 import io
 import json
 import os
 from pathlib import Path
 import shutil
+import struct
 import subprocess
 import sys
-import textwrap
+import termios
```
bmerkle: I am also testing on Windows and currently chat_session.py is Unix only unfortunately.

import fcntl, import termios are Unix-only modules

tools/chat_sessions.py
```diff
@@ -18,20 +18,98 @@
...
 from typing import Any
 
+from colorama import Fore, init, Style
+import re
```
bmerkle: re (stdlib) is imported at chat_sessions.py:33 after the third-party colorama import. It should be grouped with the other stdlib imports above.

tools/chat_sessions.py
```diff
@@ -18,20 +18,98 @@
...
+
```
bmerkle, line 33: VSCODE_USER_DIR is initialized at module level to the macOS path (line 35), then conditionally overridden in main() (line 671). If any function (e.g. load_all_sessions()) is called outside of main(), it silently uses the wrong path. Consider computing it once based on sys.platform at module level, or passing it as a parameter.

tools/chat_sessions.py
bmerkle, line 205: pads gaps with None, which means raw_requests can contain None entries

bmerkle: is catching all Exceptions a good idea ? e.g.  JSON errors vs. other Errors like OSError

tools/chat_sessions.py
```diff
@@ -344,36 +458,71 @@ def show_session(session: SessionInfo) -> None:
...
+            elif tools and not cancelled:
+                print(
+                    f"\nCOPILOT ({model_short}): ({len(tools)} tool call(s), no text response)"
+                )
 
         print()
 
 
-def search_sessions(sessions: list[SessionInfo], query: str) -> None:
+def search_sessions(
```
bmerkle, line 801: only searches user and assistant text, skipping thinking and tool call content. 
May be intentional but worth noting

tools/chat_sessions.py
bmerkle: Shell=True  with a string is a shell injection vector — but the practical risk here is low, because everything is currently in our own control.

maybe prefer shell=False and 

import shlex

proc = subprocess.Popen(
    shlex.split(pager_cmd),
    stdin=subprocess.PIPE,
    encoding="utf-8",
    errors="replace",
    env=env,
)

tools/chat_sessions.py
```diff
@@ -382,20 +531,53 @@ def search_sessions(sessions: list[SessionInfo], query: str) -> None:
...
+                        )  # -10 for ANSI codes
+                        half_avail = available // 2
+                        start = max(0, idx - half_avail)
+                        end = min(len(text), idx + len(query) + half_avail)
+                        # Redistribute unused budget when one side hits a boundary
+                        if start == 0 and end < len(text):
+                            end = min(len(text), end + (half_avail - idx))
+                        elif end == len(text) and start > 0:
+                            start = max(
+                                0, start - (half_avail - (len(text) - idx - len(query)))
```
bmerkle: Snippet budget redistribution can go negative. If idx is very large and the match is near the end, the expression half_avail - (len(text) - idx - len(query)) at line 557 can produce a negative value, making start overshoot backward. max(0, ...) guards against going below zero, but the intent is muddled. 
A simpler approach: compute start and end once, then expand whichever side has room.

bmerkle (commented):
@gvanrossum here is a first review. 
currently the tools in Unix only unfortunately

gvanrossum:
Thanks Bernhard! I will get the agent working on this.

gvanrossum:
@bmerkle -- please re-review. I have done some more development and addressed all your concerns (plus some of my own).

It was an interesting journey -- e.g. I discovered that Windows 'more' doesn't understand ANSI escapes and counts them as regular characters, and that totally messes up, notably on narrower terminal windows. (Claude was no help here!)

bmerkle (approved):
works really great now
LGTM, approved

PR #235, opened 2026-04-10
perf: Replace black with stdlib pprint for runtime formatting

KRRT7:
Follows up on #229 (defer black import). Guido [suggested](https://github.com/microsoft/typeagent-py/pull/229#issuecomment-4226329262) making `black` fully optional at runtime.

---

`black` was only used at runtime in two cold formatting paths:

- **`knowpro/answers.py`** — `create_context_prompt()` formats a dict for LLM debug context
- **`aitools/utils.py`** — `format_code()` / `pretty_print()` formats Python literals for terminal output

Both format Python data structures, which is exactly what `pprint.pformat` does. Replaced `black.format_str` with `pprint.pformat` + `ast.literal_eval` — stdlib only, no conditional imports, no try/except.

Moved `black` from `dependencies` to `dev` dependency-group. It remains available for `make format` / `make check` but is no longer required by library consumers.

## Changes

- `answers.py`: `pprint.pformat(dictify(context), width=200)` replaces `black.format_str(str(dictify(context)), ...)`
- `utils.py`: `pretty_print` uses `pprint.pformat` directly; `format_code` uses `ast.literal_eval` + `pprint.pformat`
- `pyproject.toml`: `black` moved from `dependencies` to `dev` dependency-group
- `test_utils.py`: updated existing test, added 3 new tests for `pretty_print` and `format_code`

|:---|---:|---:|---:|
| `import typeagent` | 713 ms ± 8 ms | 713 ms ± 8 ms | — |

Import time is unchanged because #229 already deferred the `import black` to first use. This PR removes the deferred import entirely and drops `black` + transitive deps (`pathspec`, `platformdirs`, `tomli`) from the runtime dependency tree.

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ classifiers = [
 ]
 dependencies = [
   "azure-identity>=1.22.0",
-  "black>=25.12.0",
   "colorama>=0.4.6",
   "mcp[cli]>=1.12.1",
   "numpy>=2.2.6",
@@ -81,6 +80,7 @@ known_local_folder = ["conftest"]
 dev = [
   "azure-mgmt-authorization>=4.0.0",
   "azure-mgmt-keyvault>=12.1.1",
+  "black>=25.12.0",
   "coverage[toml]>=7.9.1",
   "google-api-python-client>=2.184.0",
   "google-auth-httplib2>=0.2.0",
--- a/src/typeagent/aitools/utils.py
+++ b/src/typeagent/aitools/utils.py
@@ -44,27 +44,24 @@ def timelog(label: str, verbose: bool = True):
 
 
 def pretty_print(obj: object, prefix: str = "", suffix: str = "") -> None:
-    """Pretty-print an object using black.
+    """Pretty-print an object using pprint."""
+    import pprint
 
-    NOTE: Only works if its repr() is a valid Python expression.
-    """
-    print(prefix + format_code(repr(obj)) + suffix)
+    line_width = min(200, shutil.get_terminal_size().columns)
+    print(pprint.pformat(obj, width=line_width))
 
 
 def format_code(text: str, line_width=None) -> str:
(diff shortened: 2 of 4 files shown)


gvanrossum (approved):

gvanrossum:
@KRRT7 I like what this does, so I'm approving it. (**EDIT:** And merging.) Nevertheless I'd like to know how much you thought about this and how much was done by a coding agent (and which one). Could you be upfront about that please? Or *are* you an agent?

KRRT7:
I'm human :) We actually met briefly at PyCon last year — I mentioned I was building an optimization agent (codeflash) and you gave some helpful feedback. At PyBay when you demo'd typeagent it caught my attention because I thought I had an application for it, but got too busy to keep implementing it — anyways.

I've been building the agent for the past several months and I'm dogfooding it on open source projects to stress-test it on real codebases. For this PR specifically: I pointed the agent to [your comment](https://github.com/microsoft/typeagent-py/pull/229#issuecomment-4226329262) (GH MCP) about conditionally importing black, and it did the initial profiling and initially implemented a try/except import wrapper — catching `ModuleNotFoundError` and moving black to the optional dependency group. I reviewed that and didn't like it.

From the session transcript, I told it: *"we want to come up with a way where 1) we don't do try except 2) move black to optional dependencies and 3) get the correct result"* and *"can we somehow get equivalent results with just stdlib or hook into another library somehow or similar"*. The agent found `pprint.pformat` as the stdlib replacement and implemented it.

So the split is roughly: agent did profiling + implementation, I reviewed everything and directed the design decisions.

gvanrossum:
Cool. I wish you had mentioned that in the initial PR comment.

Worse, this introduced a bug that broker eval runs. I have a fix in #237.

Not sure if that's due to your optimization agent.

KRRT7:
Yes, it's on us, thanks for letting me know, I'll try to be more transparent about things in the future.

PR #236, opened 2026-04-10
perf: Defer query-time imports in conversation_base

KRRT7:
**Stack: depends on #235.** Merge #235 first, then this PR.

---

`conversation_base.py` eagerly imported four query-time modules at module level:
- `answers` (20 ms cumulative — pulls in `search` + `query`)
- `search_query_schema` (10 ms)
- `searchlang` (3 ms)
- `answer_response_schema` (1 ms)

These are only used in the `query()` method. Moved the imports inside `query()` and used `TYPE_CHECKING` + `from __future__ import annotations` for the type hints.

| Benchmark | Before (#235) | After | Speedup |
|:---|---:|---:|---:|
| `import typeagent` | 734 ms ± 13 ms | 683 ms ± 15 ms | **1.07x** |

Cumulative since baseline (main):

| Benchmark | main | This PR | Speedup |
|:---|---:|---:|---:|
| `import typeagent` | 791 ms ± 11 ms | 683 ms ± 15 ms | **1.16x** |

---

### Upstream opportunity: pydantic_ai transitive imports

`pydantic_ai` contributes ~161 ms cumulative to `import typeagent` through its transitive dependency chain:

| Module | Cumulative (ms) | Notes |
|:---|---:|:---|
| `logfire` | 43 | Observability SDK, imported eagerly |
| `griffe` | 34 | Code introspection for tool schemas |
| `pydantic_ai.messages` | 22 | Pydantic model definitions |
| `genai_prices` | 9 | Pricing data snapshot |

This is the single largest remaining import cost and would need to be addressed upstream in pydantic-ai.

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ classifiers = [
 ]
 dependencies = [
   "azure-identity>=1.22.0",
-  "black>=25.12.0",
   "colorama>=0.4.6",
   "mcp[cli]>=1.12.1",
   "numpy>=2.2.6",
@@ -81,6 +80,7 @@ known_local_folder = ["conftest"]
 dev = [
   "azure-mgmt-authorization>=4.0.0",
   "azure-mgmt-keyvault>=12.1.1",
+  "black>=25.12.0",
   "coverage[toml]>=7.9.1",
   "google-api-python-client>=2.184.0",
   "google-auth-httplib2>=0.2.0",
--- a/src/typeagent/aitools/utils.py
+++ b/src/typeagent/aitools/utils.py
@@ -44,27 +44,24 @@ def timelog(label: str, verbose: bool = True):
 
 
 def pretty_print(obj: object, prefix: str = "", suffix: str = "") -> None:
-    """Pretty-print an object using black.
+    """Pretty-print an object using pprint."""
+    import pprint
 
-    NOTE: Only works if its repr() is a valid Python expression.
-    """
-    print(prefix + format_code(repr(obj)) + suffix)
+    line_width = min(200, shutil.get_terminal_size().columns)
+    print(pprint.pformat(obj, width=line_width))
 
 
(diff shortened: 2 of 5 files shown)


gvanrossum (commented):
I don't follow precisely what games you are playing with `from __future__ import annotations`, but I don't want it in my code -- while it's not yet deprecated, eventually it will be, and I don't want to add a future burden. If that means the whole optimization can't happen, so be it -- as I wrote before, these optimizations aren't all that valuable given the typical LLM roundtrip.

KRRT7 (commented):
Good point — I can drop `from __future__ import annotations` and use string literals for the four annotations that reference the deferred modules instead:

```python
_query_translator: "typechat.TypeChatJsonTranslator[search_query_schema.SearchQuery] | None" = None
_answer_translator: "typechat.TypeChatJsonTranslator[answer_response_schema.AnswerResponse] | None" = None
```

and in the `query()` signature:

```python
async def query(
    self,
    question: str,
    search_options: "searchlang.LanguageSearchOptions | None" = None,
    answer_options: "answers.AnswerContextOptions | None" = None,
) -> str:
```

The `TYPE_CHECKING` block + runtime import inside `query()` stay the same — only the future import goes away. Type checkers still resolve the string annotations against the `TYPE_CHECKING` imports. I'll push that update.

On the value question — I was honestly hesitant to open import-time optimizations at all, which is why I [asked first on #229](https://github.com/microsoft/typeagent-py/pull/229#issuecomment-4225624953) before going further. That said, if you'd rather keep the code straightforward and not defer these, I'm happy to close this one.

gvanrossum:
Sorry, on second thought I do indeed think this isn't worth it. When querying these modules are needed. When ingesting, we save maybe 34 msec, whereas the total ingestion time even for a single message is measured in seconds.

PR #237, opened 2026-04-12
Add 'make eval'. Fix crash in pprint(). Reformat some files.

gvanrossum:
Apparently at least one of isort and black was updated and now produces more compact code in some cases.

Also change the black call in Makefile to use only -tpy312 since black on 3.12 cannot parse back the py314 code it generated (maybe only the last -t flag was effective?).

And finally make Makefile more robust by using `uv run X` instead of `.venv/bin/X`.

### DIFF
--- a/src/typeagent/aitools/utils.py
+++ b/src/typeagent/aitools/utils.py
@@ -55,13 +55,18 @@ def format_code(text: str, line_width=None) -> str:
     """Format a Python literal expression using pprint.
 
     NOTE: The text must be a valid Python literal expression (as produced by repr()).
+    Falls back to plain text formatting if the text is not a valid literal.
     """
     import ast
     import pprint
 
     if line_width is None:
         line_width = min(200, shutil.get_terminal_size().columns)
-    return pprint.pformat(ast.literal_eval(text), width=line_width)
+    try:
+        return pprint.pformat(ast.literal_eval(text), width=line_width)
+    except (ValueError, SyntaxError):
+        # Fall back to simple pprint of the string itself if it's not a valid literal
+        return pprint.pformat(text, width=line_width)
 
 
 def reindent(text: str) -> str:
--- a/src/typeagent/aitools/utils.py
+++ b/src/typeagent/aitools/utils.py
@@ -48,7 +48,7 @@ def pretty_print(obj: object, prefix: str = "", suffix: str = "") -> None:
     import pprint
 
     line_width = min(200, shutil.get_terminal_size().columns)
-    print(pprint.pformat(obj, width=line_width))
+    print(prefix + pprint.pformat(obj, width=line_width) + suffix)
 
 
 def format_code(text: str, line_width=None) -> str:
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ format: venv
 
 .PHONY: check
 check: venv
-	uv run pyright --pythonpath python src tests tools examples
+	uv run pyright src tests tools examples
 
 .PHONY: test
 test: venv
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,12 @@ demo: venv
 compare: venv
 	.venv/bin/python -m tools.query --batch $(FLAGS)
 
+.PHONY: eval
+eval: venv
+	rm -f eval.db
+	.venv/bin/python tools/load_json.py --database eval.db tests/testdata/Episode_53_AdrianTchaikovsky_index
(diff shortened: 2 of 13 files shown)


gvanrossum-ms (approved):

src/typeagent/aitools/utils.py
```diff
@@ -55,13 +55,18 @@ def format_code(text: str, line_width=None) -> str:
...
         line_width = min(200, shutil.get_terminal_size().columns)
-    return pprint.pformat(ast.literal_eval(text), width=line_width)
+    try:
+        return pprint.pformat(ast.literal_eval(text), width=line_width)
+    except (ValueError, SyntaxError):
+        # Fall back to simple pprint of the string itself if it's not a valid literal
+        return pprint.pformat(text, width=line_width)
 
 
 def reindent(text: str) -> str:
```
KRRT7, line 72: `pprint.pformat(text)` on a plain string returns its `repr()` — e.g. `'<CustomClass object at 0x...>'` with surrounding quotes. Since `text` is already a repr string intended for terminal/diff output, returning it directly avoids the extra quoting:

```suggestion
    except (ValueError, SyntaxError):
        # Fall back to the raw text if it's not a valid literal
        return text
```

KRRT7 (commented):
Thanks for catching the `format_code` crash — that's on us from #235. Two notes:

1. **`format_code` fallback wraps text in quotes** — `pprint.pformat(text)` on a string returns its `repr()`, so a non-literal like `<CustomClass object at 0x...>` becomes `'<CustomClass object at 0x...>'` with surrounding quotes in the output. Since `text` is already a repr string meant for display, returning it directly would be cleaner (inline comment below).

2. **`pretty_print` silently drops `prefix`/`suffix`** — This is a separate regression from #235 that isn't addressed here. Our PR changed the body to `print(pprint.pformat(obj, width=line_width))`, dropping the `prefix`/`suffix` args. Callers like `utils.pretty_print(actual1, Fore.GREEN, Fore.RESET)` lost their color output. Should be `print(prefix + pprint.pformat(obj, width=line_width) + suffix)`. Happy to send a follow-up PR for this if you'd like.

gvanrossum:
Ah, thanks. If you add the fixes to this PR using the GitHub suggestion feature I will apply them.

KRRT7:
The `format_code` suggestion is up as an inline comment. The `pretty_print` fix can't be a GitHub suggestion since it's not in this PR's diff — here's the change for copy-paste if you'd like to include it:

In `src/typeagent/aitools/utils.py`, `pretty_print` currently ignores `prefix`/`suffix`:

```python
def pretty_print(obj: object, prefix: str = "", suffix: str = "") -> None:
    """Pretty-print an object using pprint."""
    import pprint

    line_width = min(200, shutil.get_terminal_size().columns)
    print(pprint.pformat(obj, width=line_width))
```

Should be:

```python
def pretty_print(obj: object, prefix: str = "", suffix: str = "") -> None:
    """Pretty-print an object using pprint."""
    import pprint

    line_width = min(200, shutil.get_terminal_size().columns)
    print(prefix + pprint.pformat(obj, width=line_width) + suffix)
```

Otherwise happy to send a separate PR for it.

src/typeagent/aitools/utils.py
```diff
@@ -55,13 +55,18 @@ def format_code(text: str, line_width=None) -> str:
...
     import pprint
 
     if line_width is None:
         line_width = min(200, shutil.get_terminal_size().columns)
-    return pprint.pformat(ast.literal_eval(text), width=line_width)
+    try:
+        return pprint.pformat(ast.literal_eval(text), width=line_width)
+    except (ValueError, SyntaxError):
+        # Fall back to simple pprint of the string itself if it's not a valid literal
+        return pprint.pformat(text, width=line_width)
```
KRRT7, line 69:
```suggestion
        return text
```
`pprint.pformat` on a plain string wraps it in quotes — returning `text` directly avoids that.

KRRT7:
Opened https://github.com/gvanrossum/typeagent-py/pull/1 against your `evals` branch with the `pretty_print` prefix/suffix fix — one-click merge if you want it.

src/typeagent/aitools/utils.py
gvanrossum: Let's put a pin in that until we see its effect. (In any case you would have to update the comment too.)

src/typeagent/aitools/utils.py
gvanrossum: That suggestion would make the code incorrect.

gvanrossum (commented):
@KRRT7 Let's just merge this as-is now -- it fixes the prefix/suffix and the rest is debatable.

bmerkle (approved):
LGTM

Issue #238, opened 2026-04-13
regression URL parsing in PR #231


bmerkle:
Hej, @KRRT7 @gvanrossum 

Seems we have a regression with the PR #231 
https://github.com/microsoft/typeagent-py/pull/231
https://github.com/microsoft/typeagent-py/commit/02ca2d2c7e8c8c3a91f43b91bc5e87a44c962dfe

My .env file contains AZURE_OPENAI_ENDPOINT and AZURE_OPENAI_ENDPOINT_EMBEDDING  which _includes_ the api version after ?
Note that this is default, how you get it from the foundry UI when creating new deployments, to this format should be the default in .env files and it should also work out of the box. (as most users will use this format)

make test now fails with my .env files containing ? and api version

 we have two problems:
1. AsyncAzureOpenAI then appends its own /openai/deployments/gpt-4o/chat/completions, resulting in a doubled path → 404.
2. The test_online.py test uses create_async_openai_client (a different code path in utils.py), whereas the failing test goes through create_chat_model → _make_azure_provider → AzureProvider(openai_client=AsyncAzureOpenAI(...)).

KRRT7:
Hi @bmerkle I'm using https://ai.azure.com and was not having the same behavior, what version of the python SDK are you on?

bmerkle:
I am on the versions, defined in the project.toml.
another thing is that .\make format yields changes (which is strange)

just creating a PR for this issue.

KRRT7:
Dug into this — the root cause is in `parse_azure_endpoint` (`src/typeagent/aitools/utils.py`).

PR #231 strips the query string (`?api-version=...`) but doesn't strip the `/openai/deployments/{name}` path that Azure Foundry includes in its default endpoint URLs. So when the endpoint from `.env` looks like:

```
https://foo.openai.azure.com/openai/deployments/gpt-4o?api-version=2024-06-01
```

After the `split("?")` fix, `parse_azure_endpoint` returns:

```
https://foo.openai.azure.com/openai/deployments/gpt-4o
```

Then `AsyncAzureOpenAI.__init__` appends `/openai` to that (since no `azure_deployment` kwarg is passed), producing:

```
https://foo.openai.azure.com/openai/deployments/gpt-4o/openai/chat/completions
```

→ 404.

Fix would be to also strip the `/openai/deployments/{name}` path (and trailing `/openai`) in `parse_azure_endpoint` after stripping the query string:

```python
[...]

PR #239, opened 2026-04-13
regression URL parsing in PR #231

bmerkle:
fixes bugs described in #238 
- regression URL parsing in PR #231
- make format on main


src/typeagent/aitools/utils.py
```diff
@@ -200,9 +200,12 @@ def parse_azure_endpoint(
...
         )
 
-    # Strip query string — AsyncAzureOpenAI expects a clean base URL and
-    # receives api_version as a separate parameter.
+    # Strip query string and deployment path — AsyncAzureOpenAI expects a
+    # clean base URL and builds the deployment path internally.
     clean_endpoint = azure_endpoint.split("?", 1)[0]
+    deployment_idx = clean_endpoint.find("/openai/deployments/")
+    if deployment_idx >= 0:
+        clean_endpoint = clean_endpoint[:deployment_idx]
```
KRRT7: This handles `/openai/deployments/{name}` but misses a bare `/openai` suffix. `AsyncAzureOpenAI` always appends `/openai` to the base URL (see `openai/lib/azure.py:226-228`), so an endpoint like `https://foo.openai.azure.com/openai?api-version=...` would produce `/openai/openai/...`.

Suggested addition after this line:
```python
# Also strip a bare /openai suffix — the SDK always appends /openai.
clean_endpoint = re.sub(r"/openai$", "", clean_endpoint)
```

src/typeagent/aitools/model_adapters.py
```diff
@@ -188,6 +188,7 @@ def _make_azure_provider(
             azure_endpoint=azure_endpoint,
             api_version=api_version,
             api_key=raw_key,
+            default_headers={"Ocp-Apim-Subscription-Key": raw_key},
```
KRRT7: This should be lifted above the `if/else` so the identity branch also gets the header. As-is, managed identity behind APIM silently drops the subscription key.

```python
apim_key = os.getenv("AZURE_APIM_SUBSCRIPTION_KEY")
default_headers = (
    {"Ocp-Apim-Subscription-Key": apim_key} if apim_key else None
)

if raw_key.lower() == "identity":
    ...
    client = AsyncAzureOpenAI(..., default_headers=default_headers)
else:
    client = AsyncAzureOpenAI(..., default_headers=default_headers)
```

KRRT7 (changes requested):
Two edge cases to address before merging:

1. **Bare `/openai` suffix not stripped** — If an endpoint looks like `https://foo.openai.azure.com/openai?api-version=...`, the current fix leaves `/openai` intact. `AsyncAzureOpenAI.__init__` unconditionally appends `/openai` to the base URL (see `openai/lib/azure.py:226-228`), producing `https://foo.openai.azure.com/openai/openai/...` → 404. Trivial fix — add a `re.sub(r"/openai$", "", clean_endpoint)` after the deployment path stripping.

2. **`AZURE_APIM_SUBSCRIPTION_KEY` header missing from the identity branch** — The APIM header is only added in the `else` (raw key) branch of `_make_azure_provider`. If someone uses managed identity behind Azure API Management, the `Ocp-Apim-Subscription-Key` header is silently omitted. Lift the `apim_key` lookup and `default_headers` construction above the `if/else` so both branches get it.

src/typeagent/aitools/model_adapters.py
```diff
@@ -188,6 +188,7 @@ def _make_azure_provider(
...
         )
```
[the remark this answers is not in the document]
bmerkle: The copilot review comment is fully resolved. No changes needed.

tools/query.py
```diff
@@ -36,13 +36,15 @@
...
-    knowledge_schema as kplib,
+)
+from typeagent.knowpro import (
     query,
     search,
     search_query_schema,
     searchlang,
     serialization,
 )
+from typeagent.knowpro import knowledge_schema as kplib
```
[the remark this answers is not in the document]
bmerkle: formatted via isort

gvanrossum:
@KRRT7 Can you review and either approve or add feedback for @bmerkle? Once you two agree, and no sooner, can I review/skim and merge.

src/typeagent/aitools/model_adapters.py
bmerkle: outdated

src/typeagent/aitools/utils.py
bmerkle: i think we should use:
```
   # Strip query string and /openai... path — AsyncAzureOpenAI expects a
    # clean base URL and builds the deployment path internally.
    clean_endpoint = azure_endpoint.split("?", 1)[0]
    clean_endpoint = re.sub(r"/openai(/deployments/.*)?$", "", clean_endpoint)

```

KRRT7 (approved):

src/typeagent/aitools/utils.py
```diff
@@ -200,12 +200,10 @@ def parse_azure_endpoint(
...
         )
 
-    # Strip query string and deployment path — AsyncAzureOpenAI expects a
+    # Strip query string and /openai... path — AsyncAzureOpenAI expects a
     # clean base URL and builds the deployment path internally.
     clean_endpoint = azure_endpoint.split("?", 1)[0]
-    deployment_idx = clean_endpoint.find("/openai/deployments/")
-    if deployment_idx >= 0:
-        clean_endpoint = clean_endpoint[:deployment_idx]
+    clean_endpoint = re.sub(r"/openai(/deployments/.*)?$", "", clean_endpoint)
```
KRRT7, line 206: I'd personally re.compile the regex at module level but that's a nitpick.

gvanrossum: You know 're' uses a cache for compilations, right?

KRRT7: yes, but explicit is better than implicit, and sometimes the cache is a perf bottleneck: https://github.com/django/django/pull/20096

gvanrossum: Clearly you have per on your mind. In this tool my claim is that it doesn't matter; and I prefer having things local rather than having to look them up in  the global scope (a distraction for the reader of the code).

KRRT7: yes, I don't disagree, which is why in this case I said it's a nitpick rather than requesting a change, I've already approved of the PR.

gvanrossum: You were pushing back mightily for a nitpick. :-)

I will review and merge.

gvanrossum (approved):

PR #249, opened 2026-04-24
Setup coverage (windows)

bmerkle:
This PR:
fixed setup for uv coverage under Windows

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,7 +3,7 @@
 [run]
 branch = true
 parallel = true
-source = typeagent
+source = src
 
 [report]
 sort = cover
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,7 +3,7 @@
 [run]
 branch = true
 parallel = true
-source = test
+source = typeagent
 
 [report]
 sort = cover
--- a/make.bat
+++ b/make.bat
@@ -13,6 +13,7 @@ if "%~1"=="" goto help
 if /I "%~1"=="format" goto format
 if /I "%~1"=="check" goto check
 if /I "%~1"=="test" goto test
+if /I "%~1"=="coverage" goto coverage
 if /I "%~1"=="demo" goto demo
 if /I "%~1"=="build" goto build
 if /I "%~1"=="venv" goto venv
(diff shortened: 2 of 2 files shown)


gvanrossum (approved):

KRRT7:
A couple of things I noticed looking at this:

1. **`source = src` should probably be `source = src,test`** — the previous `source = test` was wrong in the other direction, but this just swapped which half we're missing. Measuring test coverage catches dead test code from copy-paste naming mistakes, obsolete helpers, and tests that silently never run. Ned Batchelder has a good writeup on why: https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html

> [!IMPORTANT]
> This matters even more now that AI agents are writing and modifying tests. They're great at producing code that *looks* correct — proper structure, reasonable assertions, good names — but they regularly introduce subtle bugs: tests that pass because they assert the wrong thing, helper functions that never get called, branches in test logic that are dead on arrival. The failure mode is silent: CI is green, coverage of `src` looks fine, nobody notices the test isn't actually exercising anything. Covering your test code is a cheap defensive check against that. If an agent generates a test function but misspells the name so pytest doesn't collect it, `source = src` won't catch it — `source = src,test` will show it as uncovered immediately.

---

2. **`make.bat` coverage target is missing `COVERAGE_PROCESS_START`** — the Makefile sets it; `make.bat` doesn't. Without it, you get zero coverage from the MCP server code on Windows. Should be:

   ```bat
   set "COVERAGE_PROCESS_START=.coveragerc"
   uv run coverage run -m pytest
   set "COVERAGE_PROCESS_START="
   ```

   <details>
   <summary>Why this matters</summary>

   The MCP server tests spawn the server as a subprocess (`test_mcp_server.py` launches `python -m typeagent.mcp.server` via `StdioServerParameters`). `coverage run -m pytest` only measures the main pytest process — it can't see into child processes. `COVERAGE_PROCESS_START` is the mechanism that bridges that gap: the test fixture propagates it to the subprocess env, `server.py` calls `coverage.process_startup()` which sees the var and starts collecting, and `coverage combine` merges everything at the end. Without it, `process_startup()` is a no-op.

   </details>

bmerkle:
1. regarding the inclusion of the test directory, I would disagree. 
The coverage should report the coverage of the **production code** or **production app**. Mixing up the src and test code dilutes the number or metric.
I would also not necessarily measure coverage on infrastructure code which was only generated via a template or template based code generator. 
The argument has not changed with AI IMO, but I also asked on a mailing list where I know some TDD experts, about their opinion.... 
Also please look at the comments too the article you cited... 


2. this is a bug for subprocesses. Without it, subprocess coverage (spawned by pytest) isn't collected. 
Thanks for finding it. Created a fix #253

PR #251, opened 2026-04-25
Add benchmark tools! Part of ccc4d55 / #243

gvanrossum-ms:
These files were supposed to be moved as part of #243 but because of an operator error weren't. Luckily I still had them in my workspace.

### DIFF
--- a/tools/benchmark_query.py
+++ b/tools/benchmark_query.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Benchmark lookup_term_filtered as a standalone script.
+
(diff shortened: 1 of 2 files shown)


gvanrossum (approved):

KRRT7 (changes requested):
I have some thoughts, let me get them in order.

gvanrossum:
> I have some thoughts, let me get them in order.

The original PR, that this was supposed to be a part of, was #247. Sadly I made a copy/paste error in the merge commit message for this PR. That was in turn just moving the benchmark tests you created out of the tests directory, where they were run each time by `make test` or `pytest tests`. Since they produce output and we don't watch that in CI, I think it's better they not be run as part of regular test runs.

PR #252, opened 2026-04-25
Separate batching from concurrency during ingestion

gvanrossum-ms:
Batch size is now just something the ingestion tools can use to control transaction size/frequency(*). Concurrency is how many concurrent knowledge extraction tasks to start concurrently.

To maximize throughput if you have paid your provider to allow you to use lots of tokens/minute, use a huge batch size and set concurrency just high enough that you don't overload your provider. However, since a single failure throws away your entire batch, you are advised not to set your batch size *too* large. Tune based on experiments.

(*) My next plan is to add a new top-level API that you feed with an iterator, where you can separately control concurrency and transaction batch size. ~Not sure yet whether I'll make that part of this PR; probably not.~ Wait for those commits to come soon.

@KRRT7

### DIFF
--- a/src/typeagent/knowpro/conversation_base.py
+++ b/src/typeagent/knowpro/conversation_base.py
@@ -3,6 +3,7 @@
 
 """Base class for conversations with incremental indexing support."""
 
+from collections.abc import AsyncIterable
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Generic, Self, TypeVar
@@ -36,6 +37,8 @@ from .interfaces import (
     MessageOrdinal,
     Topic,
 )
+from .knowledge import extract_knowledge_from_text_batch
+from .messageutils import get_all_message_chunk_locations
 
 TMessage = TypeVar("TMessage", bound=IMessage)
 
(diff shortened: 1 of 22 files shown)


gvanrossum-ms:
~My laptop battery is about to die.~ Feel free to review, I'll continue later.

gvanrossum:
@KRRT7, @bmerkle: This is now ready for review. It should keep the pipeline busy continuously. See podcast_ingest.py for example usage.

tools/ingest_email.py
```diff
@@ -344,7 +344,7 @@ async def ingest_emails(
                 sys.exit(f"Authentication error: {e!r}")
 
             # Print progress periodically
-            if (success_count + failed_count) % batch_size == 0:
+            if (success_count + failed_count) % concurrency == 0:
```
bmerkle: maybe add a guard against division by zero and verify concurrency has a value != 0

tests/test_source_id_ingestion.py
```diff
@@ -0,0 +1,164 @@
...
+
+        await storage.close()
+
+
+# ---------------------------------------------------------------------------
+# Helper
+# ---------------------------------------------------------------------------
+
+
+def result_count(storage: SqliteStorageProvider) -> int:
```
bmerkle: result_count used before it is defined. 

can we move 
- result_count to the head of the file
- make it internal with _ (like _make_message) if it is used only internally

src/typeagent/knowpro/convsettings.py
```diff
@@ -54,7 +54,7 @@ def __init__(
             TextEmbeddingIndexSettings(model, min_score=0.7)
         )
         self.semantic_ref_index_settings = SemanticRefIndexSettings(
-            batch_size=4,  # Effectively max concurrency
+            concurrency=4,
```
bmerkle, line 57: possible usage inconsistency:
SemanticRefIndexSettings defaults to concurrency=4, but both add_batch_to_semantic_ref_index and add_batch_to_semantic_ref_index_from_list have a parameter default of concurrency=2

src/typeagent/storage/sqlite/provider.py
```diff
@@ -627,3 +628,56 @@ async def mark_source_ingested(
...
+
+        Idempotent: re-recording overwrites any prior entry for the same
+        (message_ordinal, chunk_ordinal). No commit; call within a transaction
+        context.
+        """
+        failed_at = datetime.now(timezone.utc).isoformat()
+        cursor = self.db.cursor()
+        cursor.execute(
+            """
+            INSERT OR REPLACE INTO ChunkFailures
```
bmerkle, line 649: existing databases possibly fail, because with schema error, because they did not have the table.
Either add a migration step (similar to how INGESTED_SOURCES_SCHEMA may have been added) or document that a fresh DB is required.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -186,6 +199,184 @@ async def add_messages_with_indexing(
...
+    ) -> AddMessagesResult:
+        """Process and commit a single batch within a transaction.
+
+        Messages whose ``source_id`` is already ingested are filtered out.
+        Extraction ``Failure``\\s are recorded as chunk failures.
+        """
+        # Filter out already-ingested sources
+        filtered: list[TMessage] = []
+        for msg in batch:
+            if msg.source_id is not None and await storage.is_source_ingested(
```
bmerkle, line 283: is_source_ingested called outside the transaction
In a multi-process scenario two workers could both pass the check and both ingest the same source.

src/typeagent/knowpro/universal_message.py
```diff
@@ -190,6 +190,11 @@ class ConversationMessage(IMessage):
         "Tags associated with the message", default_factory=list
     )
     timestamp: str | None = None
+    source_id: str | None = None
```
bmerkle: doc format of  source_id appears between two string literals

src/typeagent/knowpro/conversation_base.py
gvanrossum: I don't believe such a scenario is valid anyway, so let's not move the check into the transaction (it will just slow down other tasks that want to write to the db).

gvanrossum (approved):

PR #253, opened 2026-04-25
set COVERAGE_PROCESS_START, which the Makefile does.

bmerkle:
Without it, subprocess coverage (spawned by pytest) isn't collected


bmerkle:
@gvanrossum please merge

gvanrossum (approved):

PR #254, opened 2026-04-27
Dispose of create_async_openai_client

gvanrossum-ms:
The only call site was in test_mcp_server.py. Switch that to import things from model_adapters.py.

### DIFF
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -25,7 +25,11 @@ from mcp.types import (
 from openai.types.chat import ChatCompletionMessageParam
 import typechat
 
-from typeagent.aitools.utils import create_async_openai_client, resolve_azure_model_name
+from typeagent.aitools.utils import (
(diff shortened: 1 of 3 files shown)


gvanrossum:
@bmerkle Can you review? Should be simple.

tests/test_mcp_server.py
```diff
@@ -61,49 +59,32 @@ async def sampling_callback(
...
 
-    # Convert response to MCP format
     return CreateMessageResult(
         role="assistant",
-        content=TextContent(
-            type="text", text=response.choices[0].message.content or ""
-        ),
-        model=response.model,
+        content=TextContent(type="text", text=text),
+        model="gpt-4o",
```
bmerkle, line 87: this is now hardcoded and not used from env or response anymore. (but not really a issue)

bmerkle (approved):
only one small comment. 
Otherwise LGTM and can be merged

PR #255, opened 2026-04-27
Use 'stamina' for retries

gvanrossum-ms:
Disable the older max_retries options for pydantic_ai; instead, wrap complete() and get_embedding[s]_nocache() in `@stamina(...)` decorators.

### DIFF
--- a/src/typeagent/aitools/model_adapters.py
+++ b/src/typeagent/aitools/model_adapters.py
@@ -58,10 +58,10 @@ from .embeddings import (
 _TRANSIENT_ERRORS = (openai.RateLimitError, openai.APIConnectionError, openai.APITimeoutError, openai.InternalServerError)
 
 DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
-    *_TRANSIENT_ERRORS
+    _TRANSIENT_ERRORS
 )
 DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(
-    *_TRANSIENT_ERRORS
+    _TRANSIENT_ERRORS
 )
 
 # ---------------------------------------------------------------------------
--- a/src/typeagent/aitools/model_adapters.py
+++ b/src/typeagent/aitools/model_adapters.py
@@ -32,6 +32,7 @@ import os
 import numpy as np
 from numpy.typing import NDArray
 import stamina
+from stamina import BoundAsyncRetryingCaller
 
 import openai
 from pydantic_ai import Embedder as _PydanticAIEmbedder
@@ -54,6 +55,13 @@ from .embeddings import (
     NormalizedEmbeddings,
 )
 
+DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
+    openai.APIError
+)
+DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(
+    openai.APIError
+)
+
 # ---------------------------------------------------------------------------
 # Chat model adapter
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


gvanrossum-ms:
@KRRT7 Have a look. This uses stamina in the two essential places. ~~Maybe we need a way to be able to configure the stamina call parameters? Or other stamina features?~~

**UPDATE:** You can now configure it via ConversationSettings.

src/typeagent/aitools/model_adapters.py
```diff
@@ -52,6 +55,13 @@
     NormalizedEmbeddings,
 )
 
+DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
+    openai.APIError
```
KRRT7: `openai.APIError` catches everything including `AuthenticationError`, `BadRequestError`, `NotFoundError`, `PermissionDeniedError` — those will never succeed on retry.

Consider targeting transient errors only:
```python
_TRANSIENT_ERRORS = (openai.RateLimitError, openai.APIConnectionError, openai.APITimeoutError, openai.InternalServerError)

DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(*_TRANSIENT_ERRORS)
DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(*_TRANSIENT_ERRORS)
```

KRRT7 (changes requested):
The retriers are bound to `openai.APIError`, which is the base class for all OpenAI errors — including `AuthenticationError`, `BadRequestError`, `NotFoundError`, `PermissionDeniedError`. Those will never succeed on retry.

Should target transient errors only, something like `(openai.RateLimitError, openai.APIConnectionError, openai.APITimeoutError, openai.InternalServerError)`.

src/typeagent/aitools/model_adapters.py
```diff
@@ -52,6 +55,13 @@
...
+)
+DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(
+    openai.APIError
+)
```
KRRT7, line 70:
```suggestion
_TRANSIENT_ERRORS = (openai.RateLimitError, openai.APIConnectionError, openai.APITimeoutError, openai.InternalServerError)

DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
    *_TRANSIENT_ERRORS
)
DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(
    *_TRANSIENT_ERRORS
)
```

src/typeagent/aitools/model_adapters.py
gvanrossum: Thanks! I knew I was missing that but forgot to circle back.

gvanrossum (approved):

gvanrossum:
@bmerkle This needs your Approval so we can merge it.

bmerkle (approved):
approved

gvanrossum:
@bmerkle Thanks!

bmerkle:
strange, there were still format and import errors on main, so i now created #262

gvanrossum:
Strange indeed. I am guessing we're (once again) running different versions of isort and/or black?

I have these:

black, 26.3.1 (compiled: yes)
isort 8.0.1

bmerkle:
hm, next time i will run local build again (as usual) just to be sure. I was in rush and just trusted the review you both did.

but out of some feeling i did it later on main and there was also this type error...
image

I have the same versions, at least on main

```
 typeagent-py  git switch -c fix-format-etc
Switched to a new branch 'fix-format-etc'
 typeagent-py  black --version
black, 26.3.1 (compiled: yes)
Python (CPython) 3.13.5
 typeagent-py  isort --version

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 8.0.1
```

gvanrossum:
I can repeat everything you see on Windows. But after merging your most recent PR #262 all is clean.

There may be an explanation but I'd rather not spend more time investigating.

PR #258, opened 2026-04-29
Scope agent context: extract conventions, enforce git restrictions via hooks

KRRT7:
## Summary

- Extracts code generation conventions from `AGENTS.md` into `.github/code-conventions.md` — only loads when writing/generating code, not every interaction
- Replaces text-based "NEVER COMMIT CODE" with a `PreToolUse` hook (`scripts/block-git-mutations.sh`) that actually blocks git mutations
- Registers the hook for both Claude Code (`.claude/settings.json`) and VS Code Copilot (`.vscode/settings.json`)
- `AGENTS.md` goes from 106 → 32 lines — only core workflow rules remain

Related discussion: #257

## How it works

The hook script receives tool invocations as JSON on stdin, checks if it's a terminal command containing a git mutation (`commit`, `add`, `push`, `reset`, `rebase`, `merge`, `cherry-pick`, `revert`, `stash`, `tag`), and returns a `deny` decision. Read-only commands (`status`, `diff`, `log`, `show`, `branch`) pass through.

## Verifying hooks

**Tests:** 39 pytest tests cover blocking mutations, allowing read-only commands, allowing non-git commands, and ignoring non-terminal tools:

```
uv run pytest tests/test_hooks.py -v
```

**Live verification:**
- Claude Code: run `/hooks` to confirm the PreToolUse hook is registered
- VS Code Copilot: check agent logs for hook execution

## Docs

- Claude Code hooks: https://docs.anthropic.com/en/docs/claude-code/hooks
- VS Code Copilot hooks: https://code.visualstudio.com/docs/copilot/customization/hooks

## Test plan

- [x] `uv run pytest tests/test_hooks.py -v` — 39/39 passing
- [ ] Live test: trigger `git commit` in Claude Code session, verify it's blocked
- [ ] Live test: trigger `git commit` in VS Code Copilot session, verify it's blocked


.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "github.copilot.chat.agent.hooks": {
```
gvanrossum, line 2: This file varies per user. We can't replace it with this. At best we could move this to README.md for humans to manually add there. (Or possibly we can add it to AGENTS.md. :-)

.gitignore
```diff
@@ -1,5 +1,6 @@
-# Editor settings
-.vscode
+# Editor settings (but .vscode/settings.json is tracked for Copilot hooks)
```
gvanrossum, line 1: So I would be against this.

.claude/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "hooks": {
```
gvanrossum, line 2: Same objection as for .vscode

.github/code-conventions.md
```diff
@@ -0,0 +1,62 @@
+When generating Python code (e.g. when translating TypeScript to Python),
```
gvanrossum, line 1: I don't like hiding things under .github either. *If* we want to go through with this (I'm still skeptical) we should collect all the agent instructions in a single top-level directory (not hidden) and explain in AGENTS.md how to use them.

.github/copilot-instructions.md
```diff
@@ -1 +1,3 @@
 Get your instructions from AGENTS.md in the repo root.
+
+For code generation and style conventions, see [code-conventions.md](code-conventions.md).
```
gvanrossum, line 3: Move this into AGENTS.md

scripts/block-git-mutations.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
gvanrossum: And what about Windows? Several people (my coworkers and @bmerkle) use Windows (not WSL). See make.bat.

scripts/block-git-mutations.sh
```diff
@@ -0,0 +1,22 @@
...
+# Blocks: git commit, git add, git push, git reset, git rebase, etc.
+
+INPUT=$(cat)
+TOOL_NAME=$(echo "$INPUT" | jq -r '.tool_name')
+TOOL_INPUT=$(echo "$INPUT" | jq -r '.tool_input')
+
+if [ "$TOOL_NAME" = "runTerminalCommand" ] || [ "$TOOL_NAME" = "Bash" ]; then
+  COMMAND=$(echo "$TOOL_INPUT" | jq -r '.command // .input // empty')
+
+  if echo "$COMMAND" | grep -qE '\bgit\s+(commit|add|push|reset|rebase|merge|cherry-pick|revert|stash|tag|am|apply)\b'; then
```
gvanrossum: This seems trivial to bypass. (E.g. in my global git config I have an alias 'ci' for 'commit'.)

It also gives false positives when e.g. shell quoting is used. Copilot in VSCode used to have a similar issue with its built-in command parsing; over time they improved their parsing of shell commands several times before users were happy.

scripts/block-git-mutations.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# PreToolUse hook: block git commands that mutate the repository.
+# Allows: git status, git diff, git log, git show, git branch (read-only)
```
gvanrossum: Note that git branch isn't always read-only -- it can create, rename, and delete branches.

AGENTS.md
```diff
@@ -2,18 +2,8 @@
 
 **NEVER use TEST_MODEL_NAME or "test" embedding model outside of test files**
 
-Never run git commands that make any changes. (`git status` and `git diff` are fine)
-
-**NEVER COMMIT CODE. Do not run `git commit` or any other git commands
-that make changes to the repository. Not even `git add`**
-
```
gvanrossum, line 11: I would keep this and use the hook script as a second defense.

AGENTS.md
```diff
@@ -2,18 +2,8 @@
...
 When moving, copying or deleting files, use the git commands: `git mv`, `git cp`, `git rm`
 
-When I ask to update AGENTS.md (even if maybe) extract a general rule from what I said
-before and update AGENTS.md (unless it's already in there -- maybe reformulate since
-it apparently didn't work). Also, when it looks like I state a general rule, add it to
-AGENTS.md. In all cases show what you added to AGENTS.md.
```
gvanrossum: Why delete this? I **want** it to learn from its mistakes. (Though it may have to point at a different file or files.)

gvanrossum (commented):
Not bad for a first draft, but I have some issues about what goes where.

KRRT7:
Pushed updates addressing your feedback. Here's what changed and where I disagree:

**Tracked settings files (`.vscode/settings.json`, `.claude/settings.json`)**

These aren't user-specific editor preferences — they're project-level hook registrations. Both tools explicitly distinguish shared vs personal config:

- VS Code: *"Workspace settings are specific to a project and can be shared across developers on a project... This makes it easy to share settings with others in a version-controlled (for example, Git) project."* ([docs](https://code.visualstudio.com/docs/getstarted/settings))
- Claude Code: `.claude/settings.json` is *"checked into source control and shared with your team"*; `.claude/settings.local.json` is *"not checked in, useful for personal preferences"* ([docs](https://code.claude.com/docs/en/settings))

The `.gitignore` uses `!.vscode/settings.json` to track only that file — everything else under `.vscode/` stays ignored.

**`.github/` location**

`.github/copilot-instructions.md` is GitHub's standard path — Copilot auto-loads it. Moving it elsewhere means Copilot won't find it. Similarly, `.github/code-conventions.md` is linked from both AGENTS.md and copilot-instructions.md so each tool follows its own standard discovery path.

The point of splitting is lazy loading. Three tools' own docs recommend this:

- Claude Code: *"Target under 200 lines per CLAUDE.md file. Longer files consume more context and reduce adherence."* ([docs](https://code.claude.com/docs/en/memory))
- GitHub Copilot: supports `.github/instructions/NAME.instructions.md` scoped to matching file paths ([docs](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot))
- Cursor: *"Keep rules under 500 lines"* and *"Split large rules into multiple, composable rules."* ([docs](https://cursor.com/docs/context/rules))

**Hook rewritten in Python (fixes Windows, bypass, `git branch`)**

Rewrote `block-git-mutations.sh` → `block_git_mutations.py`. Addresses all three concerns:

- Cross-platform via `uv run python`. VS Code hooks natively support per-OS commands (`windows`/`linux`/`osx` keys) ([docs](https://code.visualstudio.com/docs/copilot/customization/hooks))
- Allowlist approach — only known-safe subcommands pass, everything else denied by default
- `shlex.split()` for proper shell parsing, resolves git aliases via `git config --get-regexp`, handles `&&`/`||`/`;`/`|` chains
- `git branch -d/-D/-m`, `git tag -a/-d/-f`, `git stash` (without `list`/`show`), `git worktree` (without `list`) all correctly blocked

92 tests, pure Python, no subprocess dependency.

**AGENTS.md — trimmed, not restored**

You said *"use the hook script as a second defense"* — I took that further. The hook is the *first* line of defense: deterministic, zero context cost, can't be talked out of it. That means AGENTS.md doesn't need to repeat the git rules in prose. It went from 106 → 35 lines. Everything that can be enforced by a hook *should* be, so the agent context budget is spent on things that actually need natural language (workflow, conventions, preferences).

PR #263, opened 2026-04-30
Speed up query.py --batch by parallelizing queries

gvanrossum-ms:
This sped up `time make eval` on my machine from 5m30s to 35s, a 9x speedup.

@KRRT7
@bmerkle review?


bmerkle (approved):
perfect 👍

PR #264, opened 2026-04-30
Update dev deps to latest versions

gvanrossum:
Not everything is the latest available -- Claude solved a dependency issue.

I'm not updating the runtime dependencies -- I don't want to force typeagent *users* to have the latest of everything, that may not work within their application.


gvanrossum:
@bmerkle can you Approve?

bmerkle (approved):

PR #265, opened 2026-05-01
Use streaming in ingest_email.py

gvanrossum-ms:
We now count chunks to determine batch size. Reporting is much improved.

### DIFF
--- a/src/typeagent/aitools/model_adapters.py
+++ b/src/typeagent/aitools/model_adapters.py
@@ -27,18 +27,21 @@ required environment variables.
 """
 
 from collections.abc import Sequence
+import logging
 import os
 
 import numpy as np
 from numpy.typing import NDArray
 import stamina
 from stamina import BoundAsyncRetryingCaller
+from stamina.instrumentation import RetryDetails, set_on_retry_hooks
 
 import openai
 from pydantic_ai import Embedder as _PydanticAIEmbedder
 from pydantic_ai.embeddings.base import EmbeddingModel as _PydanticAIEmbeddingModelBase
 from pydantic_ai.embeddings.result import EmbeddingResult, EmbedInputType
 from pydantic_ai.embeddings.settings import EmbeddingSettings
+from pydantic_ai.exceptions import ModelAPIError
 from pydantic_ai.messages import (
     ModelMessage,
     ModelRequest,
@@ -60,14 +63,41 @@ _TRANSIENT_ERRORS = (
     openai.APIConnectionError,
     openai.APITimeoutError,
     openai.InternalServerError,
+    ModelAPIError,
 )
 
 DEFAULT_CHAT_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
     _TRANSIENT_ERRORS
 )
-DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=4, timeout=30).on(
+DEFAULT_EMBED_RETRIER = stamina.AsyncRetryingCaller(attempts=6, timeout=120).on(
     _TRANSIENT_ERRORS
 )
+
+_logger = logging.getLogger("stamina")
+
+_CALLABLE_LABELS: dict[str, str] = {
+    "request": "chat",
+    "embed_documents": "embedding",
+}
+
+
+def _on_retry(details: RetryDetails) -> None:
+    kind = _CALLABLE_LABELS.get(details.name, details.name)
+    caused = details.caused_by
+    exc_summary = repr(caused)[:200]
+    _logger.warning(
+        "stamina: retrying %s request (attempt %d, waited %.1fs so far, "
+        "waiting %.1fs): %s",
+        kind,
+        details.retry_num,
+        details.waited_so_far,
+        details.wait_for,
+        exc_summary,
+    )
+
+
+set_on_retry_hooks([_on_retry])
+
 
 # ---------------------------------------------------------------------------
 # Chat model adapter
--- a/src/typeagent/aitools/model_adapters.py
+++ b/src/typeagent/aitools/model_adapters.py
@@ -34,7 +34,7 @@ import numpy as np
 from numpy.typing import NDArray
 import stamina
 from stamina import BoundAsyncRetryingCaller
-from stamina.instrumentation import RetryDetails, set_on_retry_hooks
+from stamina.instrumentation import get_on_retry_hooks, RetryDetails, set_on_retry_hooks
 
 import openai
 from pydantic_ai import Embedder as _PydanticAIEmbedder
@@ -96,7 +96,8 @@ def _on_retry(details: RetryDetails) -> None:
     )
 
 
-set_on_retry_hooks([_on_retry])
+if not get_on_retry_hooks():
+    set_on_retry_hooks([_on_retry])
 
 
 # ---------------------------------------------------------------------------
--- a/src/typeagent/aitools/model_adapters.py
+++ b/src/typeagent/aitools/model_adapters.py
@@ -34,7 +34,7 @@ import numpy as np
 from numpy.typing import NDArray
 import stamina
 from stamina import BoundAsyncRetryingCaller
-from stamina.instrumentation import get_on_retry_hooks, RetryDetails, set_on_retry_hooks
+from stamina.instrumentation import RetryDetails, set_on_retry_hooks
 
 import openai
 from pydantic_ai import Embedder as _PydanticAIEmbedder
@@ -96,8 +96,7 @@ def _on_retry(details: RetryDetails) -> None:
     )
 
 
-if not get_on_retry_hooks():
-    set_on_retry_hooks([_on_retry])
+set_on_retry_hooks([_on_retry])
 
 
 # ---------------------------------------------------------------------------
--- a/src/typeagent/knowpro/conversation_base.py
+++ b/src/typeagent/knowpro/conversation_base.py
@@ -269,7 +269,9 @@ class ConversationBase(
             nonlocal pending_commit, pending_skipped
             if pending_commit is not None:
                 result = await pending_commit
-                result.messages_skipped = pending_skipped
+                result = await pending_commit
+                result.messages_skipped += pending_skipped
+                _accumulate(result)
                 _accumulate(result)
                 pending_commit = None
                 pending_skipped = 0
--- a/src/typeagent/knowpro/conversation_base.py
+++ b/src/typeagent/knowpro/conversation_base.py
@@ -255,23 +255,28 @@ class ConversationBase(
             total.messages_added += result.messages_added
             total.semrefs_added += result.semrefs_added
             total.chunks_added += result.chunks_added
+            total.messages_skipped += result.messages_skipped
             if on_batch_committed:
                 on_batch_committed(result)
 
         pending_commit: asyncio.Task[AddMessagesResult] | None = None
+        pending_skipped: int = 0
 
         async def _drain_commit() -> None:
-            nonlocal pending_commit
+            nonlocal pending_commit, pending_skipped
             if pending_commit is not None:
-                _accumulate(await pending_commit)
+                result = await pending_commit
+                result.messages_skipped = pending_skipped
+                _accumulate(result)
                 pending_commit = None
+                pending_skipped = 0
 
-        async def _submit_batch(filtered: list[TMessage]) -> None:
-            nonlocal pending_commit
-            if not filtered:
+        async def _submit_batch(filtered: list[TMessage], skipped: int) -> None:
+            nonlocal pending_commit, pending_skipped
+            if not filtered and not skipped:
                 return
 
-            if should_extract:
+            if filtered and should_extract:
                 next_extraction = asyncio.create_task(
                     self._extract_knowledge_for_batch(filtered)
                 )
@@ -281,6 +286,13 @@ class ConversationBase(
             # Wait for previous commit to finish (frees the DB connection)
             await _drain_commit()
 
+            if not filtered:
+                # Nothing to commit, just report skipped
+                total.messages_skipped += skipped
+                if on_batch_committed:
+                    on_batch_committed(AddMessagesResult(messages_skipped=skipped))
+                return
+
             # Await extraction result for this batch
             extraction = await next_extraction if next_extraction is not None else None
 
@@ -289,19 +301,20 @@ class ConversationBase(
             pending_commit = asyncio.create_task(
                 self._commit_batch_streaming(storage, filtered, extraction)
             )
+            pending_skipped = skipped
 
         try:
             batch: list[TMessage] = []
             async for msg in messages:
                 batch.append(msg)
                 if len(batch) >= batch_size:
-                    filtered = await self._filter_ingested(storage, batch)
-                    await _submit_batch(filtered)
+                    filtered, skipped = await self._filter_ingested(storage, batch)
+                    await _submit_batch(filtered, skipped)
                     batch = []
 
             if batch:
-                filtered = await self._filter_ingested(storage, batch)
-                await _submit_batch(filtered)
+                filtered, skipped = await self._filter_ingested(storage, batch)
+                await _submit_batch(filtered, skipped)
 
             await _drain_commit()
         except BaseException:
@@ -317,23 +330,26 @@ class ConversationBase(
         self,
         storage: IStorageProvider[TMessage],
         batch: list[TMessage],
-    ) -> list[TMessage]:
+    ) -> tuple[list[TMessage], int]:
         """Filter out messages whose source_id has already been ingested.
 
-        Safe to call while a pending_commit task exists: is_source_ingested
-        is a synchronous SELECT on SQLite's single connection, so it won't
-        interleave with the commit task's cursor operations in asyncio's
-        cooperative model.  If the storage provider becomes truly async
-        (e.g. aiosqlite), this assumption needs revisiting.
+        Returns (filtered_messages, skipped_count).
+
+        Uses a single batch query instead of per-message lookups.
         """
+        source_ids = [m.source_id for m in batch if m.source_id is not None]
+        if source_ids:
+            ingested = await storage.are_sources_ingested(source_ids)
+        else:
+            ingested = set[str]()
         filtered: list[TMessage] = []
+        skipped = 0
         for msg in batch:
-            if msg.source_id is not None and await storage.is_source_ingested(
-                msg.source_id
-            ):
+            if msg.source_id is not None and msg.source_id in ingested:
+                skipped += 1
                 continue
             filtered.append(msg)
-        return filtered
+        return filtered, skipped
 
     async def _extract_knowledge_for_batch(
         self,
--- a/src/typeagent/knowpro/conversation_base.py
+++ b/src/typeagent/knowpro/conversation_base.py
@@ -238,7 +238,10 @@ class ConversationBase(
 
         Args:
             messages: An async iterable of messages to ingest.
-            batch_size: Number of messages per commit batch.
+            batch_size: Target number of text chunks per commit batch.
+                Messages are never split across batches, so the actual
+                chunk count may exceed ``batch_size`` if a single message
+                has more chunks than that.
             on_batch_committed: Optional callback invoked after each batch is
                 committed, receiving the batch's ``AddMessagesResult``.
 
@@ -305,12 +308,21 @@ class ConversationBase(
 
         try:
             batch: list[TMessage] = []
+            batch_chunks = 0
             async for msg in messages:
-                batch.append(msg)
-                if len(batch) >= batch_size:
+                msg_chunks = len(msg.text_chunks)
+                if batch and batch_chunks + msg_chunks > batch_size:
                     filtered, skipped = await self._filter_ingested(storage, batch)
                     await _submit_batch(filtered, skipped)
                     batch = []
+                    batch_chunks = 0
+                batch.append(msg)
+                batch_chunks += msg_chunks
+                if batch_chunks >= batch_size:
+                    filtered, skipped = await self._filter_ingested(storage, batch)
+                    await _submit_batch(filtered, skipped)
+                    batch = []
+                    batch_chunks = 0
 
             if batch:
                 filtered, skipped = await self._filter_ingested(storage, batch)
--- a/src/typeagent/knowpro/conversation_base.py
+++ b/src/typeagent/knowpro/conversation_base.py
@@ -269,9 +269,7 @@ class ConversationBase(
             nonlocal pending_commit, pending_skipped
             if pending_commit is not None:
                 result = await pending_commit
-                result = await pending_commit
                 result.messages_skipped += pending_skipped
-                _accumulate(result)
                 _accumulate(result)
                 pending_commit = None
                 pending_skipped = 0
--- a/src/typeagent/storage/sqlite/provider.py
+++ b/src/typeagent/storage/sqlite/provider.py
@@ -564,6 +564,19 @@ class SqliteStorageProvider[TMessage: interfaces.IMessage](
         row = cursor.fetchone()
         return row is not None and row[0] == STATUS_INGESTED
 
+    async def are_sources_ingested(self, source_ids: list[str]) -> set[str]:
+        """Return the subset of source_ids that have already been ingested."""
+        if not source_ids:
+            return set()
+        cursor = self.db.cursor()
+        placeholders = ",".join("?" for _ in source_ids)
+        cursor.execute(
+            f"SELECT source_id FROM IngestedSources"
+            f" WHERE source_id IN ({placeholders}) AND status = ?",
+            [*source_ids, STATUS_INGESTED],
+        )
+        return {row[0] for row in cursor.fetchall()}
+
     async def get_source_status(self, source_id: str) -> str | None:
         """Get the ingestion status of a source.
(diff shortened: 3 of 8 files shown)


gvanrossum-ms:
@KRRT7 What did I miss?
@bmerkle Approval?

KRRT7:
Here's what I caught (confirmed via `coverage.py` on all 32 streaming tests — all pass):

1. **Breaking `source_id` format change** — `str(email_file)` changed to `str(email_file.resolve())`. Existing databases ingested with relative paths won't recognize re-ingestion attempts with absolute paths, causing all previously ingested emails to be duplicated. Needs to be documented prominently and possibly a migration path.

2. **SQLite `IN (...)` variable limit** — `are_sources_ingested()` passes all source_ids as bind params. With small emails and large `batch_size`, this could exceed SQLite's `SQLITE_MAX_VARIABLE_NUMBER` (999 on older versions). Consider chunking the query.

3. **Lost `AuthenticationError` handling** — The old code had a clean `sys.exit()` for wrong API keys (confirmed: `except openai.AuthenticationError` was removed in the diff). Now it surfaces as an `ExceptionGroup`, which is a UX regression.

4. **Lost failed-source DB tracking** — Old code called `mark_source_ingested(source_id, exc_name)` on failures so re-runs skip broken files (confirmed: removed in the diff). New code only counts failures in memory, meaning broken files get re-attempted every run.

5. **Double dedup check** — `_email_generator` calls `is_source_ingested()` per email, then `_filter_ingested()` checks again per batch. The first avoids parsing already-ingested files, the second is transactional dedup — both paths exercised by coverage. Worth a comment explaining why both exist.

6. **`set_on_retry_hooks([_on_retry])` at import time** — Global side effect that replaces any previously set hooks. Coverage confirms `_on_retry` (lines 85-88) is **never exercised** by any test — entirely untested code path.

7. **Fragile `messages_skipped` mutation** — `_drain_commit()` overwrites `result.messages_skipped` with `=` instead of `+=`. If `_commit_batch_streaming` ever sets `messages_skipped` itself, it gets silently clobbered. Also, coverage confirms the all-skipped-after-pending-commit branch (lines 294-296 of `conversation_base.py`) is untested.

Posted a formal review with ````suggestion` blocks for each item.

tools/ingest_email.py
```diff
@@ -239,7 +271,7 @@ def _iter_emails(
     sliced_total = len(email_files)
     for i, email_file in enumerate(email_files):
         label = f"[{i + 1}/{sliced_total}] {email_file}"
-        yield str(email_file), email_file, label
+        yield str(email_file.resolve()), email_file, label
```
KRRT7, line 274: **Breaking change: `source_id` format changed from relative to absolute path.**

`str(email_file.resolve())` produces absolute paths, but existing databases were ingested with `str(email_file)` (relative). Re-running against an existing DB will not recognize previously ingested files and **duplicate every email**.

```suggestion
        yield str(email_file), email_file, label
```

If absolute paths are intentional, this needs a migration story (e.g. a one-time UPDATE on IngestedSources).

tools/ingest_email.py
```diff
@@ -267,6 +299,101 @@ def _print_email_verbose(email: EmailMessage) -> None:
...
+    *counters* is mutated in place to track ``parsed``, ``skipped``,
+    and ``failed`` counts for the caller's summary.
+    """
+    skip_first: str | None = None
+    skip_last: str | None = None
+    skip_count = 0
+
+    for source_id, email_file, label in _iter_emails(eml_paths, verbose, offset, limit):
+        # Check source_id before opening the file
+        if await storage.is_source_ingested(source_id):
```
KRRT7, line 348: **Double dedup: generator checks `is_source_ingested()` per-email, then `_filter_ingested()` batch-checks again.**

The generator check avoids parsing already-ingested files (cheap skip before `import_email_from_file`). The batch check in `_filter_ingested` is transactional dedup against concurrent commits. Both are useful, but worth a comment explaining the two distinct roles:

```suggestion
        # Pre-parse dedup: skip before opening the file.
        # A second dedup pass happens in _filter_ingested() to catch
        # sources committed by an earlier batch in the same run.
        if await storage.is_source_ingested(source_id):
```

src/typeagent/storage/sqlite/provider.py
```diff
@@ -564,6 +564,19 @@ async def is_source_ingested(self, source_id: str) -> bool:
         row = cursor.fetchone()
         return row is not None and row[0] == STATUS_INGESTED
 
+    async def are_sources_ingested(self, source_ids: list[str]) -> set[str]:
+        """Return the subset of source_ids that have already been ingested."""
+        if not source_ids:
+            return set()
+        cursor = self.db.cursor()
+        placeholders = ",".join("?" for _ in source_ids)
```
KRRT7: **SQLite bind variable limit.** `SQLITE_MAX_VARIABLE_NUMBER` defaults to 999 on older SQLite builds (raised to 32766 in 3.32.0+). With small emails and large `--batch-size`, `source_ids` could exceed this.

Consider chunking:

```suggestion
        placeholders = ",".join("?" for _ in source_ids)
        # SQLite's SQLITE_MAX_VARIABLE_NUMBER is 999 on older builds.
        # Chunk if needed for very large batches.
```

Or chunk the query into groups of e.g. 500:
```python
result = set()
for i in range(0, len(source_ids), 500):
    chunk = source_ids[i:i+500]
    placeholders = ",".join("?" for _ in chunk)
    cursor.execute(
        f"SELECT source_id FROM IngestedSources"
        f" WHERE source_id IN ({placeholders}) AND status = ?",
        [*chunk, STATUS_INGESTED],
    )
    result.update(row[0] for row in cursor.fetchall())
return result
```

src/typeagent/aitools/model_adapters.py
```diff
@@ -60,15 +63,42 @@
...
+        "waiting %.1fs): %s",
+        kind,
+        details.retry_num,
+        details.waited_so_far,
+        details.wait_for,
+        exc_summary,
+    )
+
+
+set_on_retry_hooks([_on_retry])
```
KRRT7, line 99: **Global side effect at import time.** This replaces any previously registered hooks. Fine for the CLI, but will silently stomp hooks in tests or library consumers that register their own.

Coverage confirms `_on_retry` (lines 85-88) is never exercised by any test — so this code path is entirely untested.

```suggestion
if not stamina.instrumentation.get_on_retry_hooks():
    set_on_retry_hooks([_on_retry])
```

(Or make it additive rather than replacing.)

src/typeagent/knowpro/conversation_base.py
```diff
@@ -255,23 +258,28 @@ def _accumulate(result: AddMessagesResult) -> None:
...
         pending_commit: asyncio.Task[AddMessagesResult] | None = None
+        pending_skipped: int = 0
 
         async def _drain_commit() -> None:
-            nonlocal pending_commit
+            nonlocal pending_commit, pending_skipped
             if pending_commit is not None:
-                _accumulate(await pending_commit)
+                result = await pending_commit
+                result.messages_skipped = pending_skipped
```
KRRT7: **Fragile `messages_skipped` mutation.** `_drain_commit()` overwrites `result.messages_skipped` after the fact. If `_commit_batch_streaming` ever starts setting `messages_skipped` itself, it gets silently clobbered.

Cleaner to pass `skipped` into the commit or construct a combined result:

```suggestion
                result = await pending_commit
                result.messages_skipped += pending_skipped
                _accumulate(result)
```

Using `+=` instead of `=` makes this additive — safe whether `_commit_batch_streaming` returns 0 or starts returning its own count.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -281,6 +289,13 @@ async def _submit_batch(filtered: list[TMessage]) -> None:
             # Wait for previous commit to finish (frees the DB connection)
             await _drain_commit()
 
+            if not filtered:
+                # Nothing to commit, just report skipped
+                total.messages_skipped += skipped
+                if on_batch_committed:
+                    on_batch_committed(AddMessagesResult(messages_skipped=skipped))
```
KRRT7, line 296: **Coverage confirms this branch (all-skipped + pending commit) is untested.** Lines 294-296 are not exercised by any test. Consider adding a test where batch N has real messages and batch N+1 is all-skipped, so this code path gets covered.

KRRT7 (changes requested):
Ran `coverage.py` against the streaming tests (32/32 pass). Coverage on the key files: `conversation_base.py` 78%, `sqlite/provider.py` 57%, `model_adapters.py` 46%. Below are findings with suggested fixes — most can be applied directly.

src/typeagent/storage/sqlite/provider.py
KRRT7: Another option that avoids the bind variable limit entirely — use `executemany` to bulk-insert into a temp table and JOIN:

```python
async def are_sources_ingested(self, source_ids: list[str]) -> set[str]:
    """Return the subset of source_ids that have already been ingested."""
    if not source_ids:
        return set()
    cursor = self.db.cursor()
    cursor.execute("CREATE TEMP TABLE _check_ids (source_id TEXT)")
    cursor.executemany(
        "INSERT INTO _check_ids VALUES (?)",
        [(sid,) for sid in source_ids],
    )
    cursor.execute(
        "SELECT i.source_id FROM IngestedSources i"
        " JOIN _check_ids c ON i.source_id = c.source_id"
        " WHERE i.status = ?",
        (STATUS_INGESTED,),
    )
    result = {row[0] for row in cursor.fetchall()}
    cursor.execute("DROP TABLE _check_ids")
    return result
```

This lets SQLite's query planner optimize the join and has no bind variable ceiling. Slightly more write overhead (temp table + inserts), so the chunked `IN (...)` loop is simpler for typical batch sizes — but this scales better if batches ever get large.

tools/ingest_email.py
gvanrossum: Absolute paths are intentional. This library is way too immature to care about backwards compatibility. In any case this is in a tool -- those are really outside the library that you install using pip or uv.

Nevertheless, I'll put it in the change logs and people can migrate how they want to.

src/typeagent/storage/sqlite/provider.py
gvanrossum: We'll go with the 500-sized chunks.

src/typeagent/knowpro/conversation_base.py
gvanrossum: Test added

src/typeagent/aitools/model_adapters.py
gvanrossum: Done (with slightly different imports).

gvanrossum:
@KRRT7 -- I've addressed all your feedback except for the backward compatibility (it's too early for that).

gvanrossum:
PS. Please don't use "Request changes". Leaving comments has the same effect and I don't get the nagging message in the PR.

src/typeagent/aitools/model_adapters.py
gvanrossum: Turns out this was wrong -- there is always a default hook that prints 'stamina.retry_scheduled' so the condition always fails and our hook never gets set. Reverting this.

KRRT7:
> PS. Please don't use "Request changes". Leaving comments has the same effect and I don't get the nagging message in the PR.

sounds good, do you prefer the flow where I make the exact code suggestions and you can click "Apply suggestions" instead?

KRRT7:
> @KRRT7 -- I've addressed all your feedback except for the backward compatibility (it's too early for that).

sounds good to me, just wasn't sure of your policy around that.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -255,23 +258,28 @@ def _accumulate(result: AddMessagesResult) -> None:
...
                 pending_commit = None
+                pending_skipped = 0
 
-        async def _submit_batch(filtered: list[TMessage]) -> None:
-            nonlocal pending_commit
-            if not filtered:
+        async def _submit_batch(filtered: list[TMessage], skipped: int) -> None:
+            nonlocal pending_commit, pending_skipped
+            if not filtered and not skipped:
                 return
```
KRRT7, line 280: **Dead guard.** All three call sites guarantee `batch` is non-empty before calling `_filter_ingested`, so `len(filtered) + skipped == len(batch) > 0`. This branch can never fire. The `if not filtered:` check at line 290 already handles the all-skipped case.

```suggestion

            if filtered and should_extract:
```

tools/ingest_email.py
```diff
@@ -301,97 +438,90 @@ async def ingest_emails(
...
+        "date_skipped": 0,
+        "failed": 0,
+        "ingested": 0,
+        "batches": 0,
+    }
+
+    def on_batch_committed(result: AddMessagesResult) -> None:
+        nonlocal last_batch_time
+        counters["ingested"] += result.messages_added
+        counters["batches"] += 1
```
KRRT7, line 467: **Noisy output for all-skipped batches.** When `_submit_batch` handles an all-skipped batch, this prints `+0 messages, +0 chunks, +0 semrefs | 0.0s (0.00s/chunk)`. Early return keeps the batch counter and timing accurate for real batches.

```suggestion
        counters["ingested"] += result.messages_added
        if not result.messages_added:
            return
        counters["batches"] += 1
```

tools/ingest_email.py
```diff
@@ -301,97 +438,90 @@ async def ingest_emails(
...
     print()
     if verbose:
-        print(f"Successfully imported {success_count} email(s)")
-        if skipped_count:
-            print(f"Skipped {skipped_count} already-ingested email(s)")
-        if failed_count:
-            print(f"Failed to import {failed_count} email(s)")
+        print(f"Successfully ingested {result.messages_added} email(s)")
+        if counters["skipped"]:
+            print(f"Skipped {counters['skipped']} already-ingested email(s)")
```
KRRT7, line 528: **Skip accounting gap.** `counters["skipped"]` counts generator-level pre-parse dedup, but `result.messages_skipped` (batch-level transactional dedup from `_filter_ingested`) is never displayed. The two populations are disjoint — a source caught by the generator is never yielded to the batch layer — so summing is safe.

```suggestion
        if counters["skipped"] + result.messages_skipped:
            print(f"Skipped {counters['skipped'] + result.messages_skipped} already-ingested email(s)")
```

tools/ingest_email.py
```diff
@@ -301,97 +438,90 @@ async def ingest_emails(
...
         print(
-            f"Imported {success_count} emails to {database} "
+            f"ingested {result.messages_added} emails to {database} "
             f"({semref_count} refs, {elapsed:.1f}s)"
         )
-        if skipped_count:
-            print(f"Skipped: {skipped_count} (already ingested)")
-        if failed_count:
-            print(f"Failed: {failed_count}")
+        if counters["skipped"]:
```
KRRT7, line 542: Same skip accounting fix for non-verbose output, plus capitalize "Ingested" to match the verbose branch.

```suggestion
            f"Ingested {result.messages_added} emails to {database} "
            f"({semref_count} refs, {elapsed:.1f}s)"
        )
        if counters["skipped"] + result.messages_skipped:
            print(f"Skipped: {counters['skipped'] + result.messages_skipped} (already ingested)")
```

KRRT7 (commented):
Re-reviewed the updated commits (`551d407`..`ec4085d`). All previous suggestions addressed — chunked IN clause, `+=` for `messages_skipped`, dedup comment, all-skipped test all look good.

Ran full suite locally: `make format` (no changes), `make check` (0 errors/warnings), `make test` (697 passed). Four small follow-ups below, all click-to-apply.

KRRT7:
> sounds good, do you prefer the flow where I make the exact code suggestions and you can click "Apply suggestions" instead?

demoing it here.

tools/ingest_email.py
KRRT7: re: the Capitalization, I know it's a minor detail but I care about them.

gvanrossum:
@KRRT7 -- have another look (mostly spit and polish)

KRRT7:
Reviewed the 3 new commits (`5bc5764`, `016e6a3`, `875b7ff`). Chunks/semrefs counters and ^C handling look good. `semrefs_added` over `semref_coll.size()` is the right fix.

A few things still open from the last round:

**Dead guard in `_submit_batch`** (`conversation_base.py`). `if not filtered and not skipped: return` can never fire — all call sites guarantee `len(batch) > 0` before `_filter_ingested`, so `filtered + skipped > 0` always. The `if not filtered:` block at line ~290 already handles the all-skipped case.

**All-skipped batch output.** `on_batch_committed` still prints `+0 messages, +0 chunks, +0 semrefs` when a batch is entirely deduped. Early return on `result.messages_added == 0` would keep the batch counter and timing accurate for real batches only.

**Skip accounting gap.** `counters["skipped"]` only counts generator-level pre-parse dedup. `result.messages_skipped` (batch-level transactional dedup from `_filter_ingested`) is never surfaced in the final summary. These are disjoint populations — a source caught by the generator never reaches the batch layer. The total skipped is `counters["skipped"] + total.messages_skipped`. This is also a problem on ^C where `result` is `None` and batch-level skips are lost entirely.

**`next_extraction` task leak.** In `_submit_batch`, if an exception occurs after `next_extraction = asyncio.create_task(...)` but before it's awaited, the task leaks. The `except BaseException` block only cancels `pending_commit`, not `next_extraction`.

gvanrossum:
> sounds good, do you prefer the flow where I make the exact code suggestions and you can click "Apply suggestions" instead?

Yes. Though in this case I found some issues with your suggestions -- it's still better, and you get credit.

bmerkle:
Just ping me when you are finshed...so i can merge

tools/ingest_email.py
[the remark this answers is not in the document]
gvanrossum: As I told Kevin, I don't care about backwards compatibility. It's labeled as alpha quality.

bmerkle (approved):
approved

gvanrossum:
Hm, I had wanted to respond to some of Copilot's and @KRRT7's review comments. But I'm happy that at least this much is now on main, we can iterate further in follow-up PRs.

src/typeagent/knowpro/conversation_base.py
gvanrossum: @KRRT7 I am confused by your suggestions. This one, when applied, generates invalid syntax. Maybe you are using a tool to generate them and it's got an off-by-one error or is working from an outdated file version? It's the case for all suggestions in this batch.

gvanrossum: You're right. If we ever revisit this code we can delete it. Bernhard merged it before I got a chance to look at your feedback, in part confused by the misaligned suggestions. It's harmless so I see no need to fix it in a separate PR.

tools/ingest_email.py
gvanrossum: Honestly, that behavior sounds reasonable to me, given that this should be extremely rare. I think to trigger it you would have to run two instances of import_email in parallel with the same input files, or pass the same file multiple times but in different batches. Neither of which you should do. And if you do the latter, the batch lookup of source IDs might still bite you if the file occurs twice in the same batch.

tools/ingest_email.py
gvanrossum: Weirdly the line numbers in the file listing don't match those in the suggestion (506 --> 528).

You may be right about the accounting, it's hard to follow for me TBH.

tools/ingest_email.py
gvanrossum: I do too and it was already fixed by some last-minute improvement to the output.

gvanrossum:
I already answered the first two issues. (Sadly your suggestions demo misfired.)

> **Skip accounting gap.** `counters["skipped"]` only counts generator-level pre-parse dedup. `result.messages_skipped` (batch-level transactional dedup from `_filter_ingested`) is never surfaced in the final summary. These are disjoint populations — a source caught by the generator never reaches the batch layer. The total skipped is `counters["skipped"] + total.messages_skipped`. This is also a problem on ^C where `result` is `None` and batch-level skips are lost entirely.

Can you talk me through this in more detail? I'm lost about what messages_skipped and skipped stand for at this point.

> **`next_extraction` task leak.** In `_submit_batch`, if an exception occurs after `next_extraction = asyncio.create_task(...)` but before it's awaited, the task leaks. The `except BaseException` block only cancels `pending_commit`, not `next_extraction`.

This one's new. I think it just requires the same treatment in the except block?

tools/ingest_email.py
```diff
@@ -134,6 +128,35 @@ def create_arg_parser() -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        "--max-chunks",
+        type=int,
+        default=20,
+        metavar="N",
+        help=(
+            "Maximum number of text chunks to keep per email. "
+            "Extra chunks are silently dropped. Default: 20."
+        ),
+    )
```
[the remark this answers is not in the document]
gvanrossum: This is intentional.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -238,7 +238,10 @@ async def add_messages_streaming(
 
         Args:
             messages: An async iterable of messages to ingest.
-            batch_size: Number of messages per commit batch.
+            batch_size: Target number of text chunks per commit batch.
+                Messages are never split across batches, so the actual
+                chunk count may exceed ``batch_size`` if a single message
+                has more chunks than that.
```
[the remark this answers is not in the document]
gvanrossum: `add_messages_streaming` is new in the upcoming release. No need to worry about backwards compatibility yet.

src/typeagent/aitools/model_adapters.py
```diff
@@ -60,15 +63,42 @@
...
+
+
```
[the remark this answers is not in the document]
gvanrossum: There's a default hook producing noise output that we want to override. When a user complains about this we'll figure out a solution.

tools/ingest_email.py
```diff
@@ -267,6 +299,103 @@ def _print_email_verbose(email: EmailMessage) -> None:
...
+        if max_chunks is not None and len(email.text_chunks) > max_chunks:
+            if verbose:
+                print(f"    Truncating {len(email.text_chunks)} chunks to {max_chunks}")
+            email.text_chunks = email.text_chunks[:max_chunks]
+
+        # Set source_id so streaming API handles dedup and tracking
+        email.source_id = source_id
+        counters["parsed"] += 1
+        yield email
+
```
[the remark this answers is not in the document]
gvanrossum: Good idea, though I doubt that this will be an issue for a long time, so I will ignore your suggestion for now.

gvanrossum:
Copilot didn't find anything new.

KRRT7:
Hey Guido — quick note on why those click-to-apply suggestions generated invalid syntax.

The suggestions in my second review were generated against commit `ec4085d`. You then pushed 3 more commits (`5bc5764`, `016e6a3`, `875b7ff`) that added/removed ~45 lines in `ingest_email.py`, shifting line numbers by ~22 lines. GitHub anchors suggestion blocks to specific diff positions at review submission time — they don't rebase when new commits land. So by the time you clicked "Apply", the anchors were stale and pointed to wrong lines.

The suggestion content itself was correct for `ec4085d`. This is a known GitHub limitation.

KRRT7:
> Can you talk me through this in more detail? I'm lost about what messages_skipped and skipped stand for at this point.

There are two independent dedup layers, each with its own skip counter:

**Layer 1 — Generator-level (`counters["skipped"]`).** In `_email_generator()`, each file's `source_id` is checked against the DB via `storage.is_source_ingested(source_id)` *before parsing*. If already ingested, `counters["skipped"] += 1` and the message is never yielded. This catches sources ingested in prior runs (or by earlier batches that already committed in the same run).

**Layer 2 — Batch-level (`result.messages_skipped`).** In `add_messages_streaming()`, when a batch is ready, `_filter_ingested()` does a second bulk dedup check via `storage.are_sources_ingested()`. This catches a race: if source_id X passed Layer 1 (wasn't ingested yet), then an earlier batch commits and marks X as ingested *before* this batch reaches `_filter_ingested()`, Layer 2 catches it. These skips are tracked as `messages_skipped` in `AddMessagesResult` and accumulated in `total.messages_skipped`.

**The gap:** The final summary in `ingest_email.py` only reported `counters["skipped"]` (Layer 1). Layer 2 skips were never surfaced:
- `on_batch_committed()` tracked `messages_added`, `chunks_added`, `semrefs_added` — but not `messages_skipped`
- The summary's "Skipped" line only showed `counters['skipped']`
- The true total skipped is `counters["skipped"] + result.messages_skipped`

**On ^C it's worse:** `result` is `None` (the `await email_memory.add_messages_streaming(...)` never returns), so `result.messages_skipped` is completely lost.

The fix adds a `batch_skipped` counter to `on_batch_committed` and reports the sum in the summary.

> This one's new. I think it just requires the same treatment in the except block?

Yes. The fix promotes `next_extraction` to a `nonlocal` (`pending_extraction`) and cancels it in the `except BaseException` block alongside `pending_commit`. The lifecycle:

1. `pending_extraction` is set when `_submit_batch` creates the extraction task
2. Cleared after `await next_extraction` consumes the result
3. If `_drain_commit()` raises between steps 1 and 2, `pending_extraction` is still in-flight → the except block cancels it

Both fixes are in #266.

gvanrossum:
Thanks! And merged. I never encountered that GitHub issue before but it makes sense.

PR #273, opened 2026-05-08
Upgrade python-multipart to >=0.0.27

gvanrossum:
As directed by Dependabot. But also fix a weird merge conflict (? see #41) that left pyright in the runtime deps.


gvanrossum-ms (approved):

bmerkle (approved):
reviewed and tested. LGTM.

PR #277, opened 2026-05-13
Completely rewrite add_messages_streaming

gvanrossum:
The throughput is now much higher -- e.g. with concurrency 10 and batch size 10, the Adrian podcast ingest in 40 seconds, compared to 90 on main (with the previous pipelining implementation).

A consequence of the new design is that the message index is now populated at the time messages are added -- the secondary index building no longer needs to do this.

### DIFF
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -194,9 +194,10 @@ async def _dispatcher_task[TMessage: IMessage](
                             stop_state.stop_at_message_id = new_stop
                         if stop_state.exception is None:
                             stop_state.exception = result.error
-            await result_queue.put(result)
         finally:
             sem.release()
+
+        await result_queue.put(result)
 
     async with asyncio.TaskGroup() as tg:
         while True:
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -364,7 +364,7 @@ async def _reassembler_task[TMessage: IMessage](
         staged_results.clear()
         staged_chunks = 0
 
-    async def _drain_consecutive_complete() -> None:
+    async def _drain_consecutive_complete(force: bool = False) -> None:
         nonlocal staged_chunks
         while True:
             assembly = assemblies.get(state.first_uncommitted_ordinal)
@@ -381,7 +381,7 @@ async def _reassembler_task[TMessage: IMessage](
 
             del assemblies[state.first_uncommitted_ordinal]
             state.first_uncommitted_ordinal += 1
-            await _commit_if_needed()
+            await _commit_if_needed(force)
 
     try:
         while True:
@@ -392,45 +392,43 @@ async def _reassembler_task[TMessage: IMessage](
             chunk_ordinal = item.chunk_id.chunk_ordinal
             message_id = item.chunk_id.message_ordinal
 
-            try:
-                if chunk_ordinal < 0 or chunk_ordinal >= item.chunk_count:
-                    raise RuntimeError(
-                        f"Invalid chunk ordinal: message_id={message_id}, "
-                        f"chunk_ordinal={chunk_ordinal}, chunk_count={item.chunk_count}"
-                    )
+            validation_error: str | None = None
+            assembly = assemblies.get(message_id)
+            if chunk_ordinal < 0 or chunk_ordinal >= item.chunk_count:
+                validation_error = (
+                    f"Invalid chunk ordinal: message_id={message_id}, "
+                    f"chunk_ordinal={chunk_ordinal}, chunk_count={item.chunk_count}"
+                )
+            elif assembly is None:
+                assembly = MessageAssembly[TMessage](
+                    message_id=message_id,
+                    chunk_count=item.chunk_count,
+                    message=item.message,
+                    chunks={},
+                )
+                assemblies[message_id] = assembly
+            elif assembly.chunk_count != item.chunk_count:
+                validation_error = (
+                    f"Mismatched chunk count for message: message_id={message_id}, "
+                    f"expected={assembly.chunk_count}, got={item.chunk_count}"
+                )
+            elif chunk_ordinal in assembly.chunks:
+                validation_error = (
+                    f"Duplicate chunk: message_id={message_id}, "
+                    f"chunk_ordinal={chunk_ordinal}, chunk_count={item.chunk_count}"
+                )
 
-                existing = assemblies.get(message_id)
-                if existing is None:
-                    existing = MessageAssembly[TMessage](
-                        message_id=message_id,
-                        chunk_count=item.chunk_count,
-                        message=item.message,
-                        chunks={},
-                    )
-                    assemblies[message_id] = existing
-                elif existing.chunk_count != item.chunk_count:
-                    raise RuntimeError(
-                        f"Mismatched chunk count for message: message_id={message_id}, "
-                        f"expected={existing.chunk_count}, got={item.chunk_count}"
-                    )
-
-                if chunk_ordinal in existing.chunks:
-                    raise RuntimeError(
-                        f"Duplicate chunk: message_id={message_id}, "
-                        f"chunk_ordinal={chunk_ordinal}, chunk_count={item.chunk_count}"
-                    )
-
-                existing.chunks[chunk_ordinal] = item
-            except Exception:
-                # On validation error, set stop flag and re-raise
-                # The finally block will drain and commit consecutive complete messages
+            if validation_error is not None:
                 stop_state.stop_at_message_id = min(
                     stop_state.stop_at_message_id, message_id
                 )
-                raise
+                raise RuntimeError(validation_error)
+
+            assert assembly is not None
+            assembly.chunks[chunk_ordinal] = item
 
             if item.error is not None:
-                existing.has_error = True
+                assembly.has_error = True
                 state.chunk_failures += 1
                 stop_state.stop_at_message_id = min(
                     stop_state.stop_at_message_id, message_id
@@ -439,8 +437,7 @@ async def _reassembler_task[TMessage: IMessage](
             await _drain_consecutive_complete()
     finally:
         # Always drain and commit consecutive complete messages before raising
-        await _drain_consecutive_complete()
-        await _commit_if_needed(force=True)
+        await _drain_consecutive_complete(force=True)
 
     state.buffered_messages = len(assemblies)
     return state
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -180,12 +180,7 @@ async def _dispatcher_task[TMessage: IMessage](
                     embedding_model=embedding_model,
                 )
                 if result.error is not None:
-                    if skip_failed_messages:
-                        print(
-                            f"Skipping message {work_item.chunk_id.message_ordinal} "
-                            f"due to extraction/embedding error: {result.error}"
-                        )
-                    else:
+                    if not skip_failed_messages:
                         new_stop = min(
                             stop_state.stop_at_message_id,
                             work_item.chunk_id.message_ordinal,
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -6,13 +6,18 @@
 import asyncio
 from collections.abc import AsyncIterable, Awaitable, Callable
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 import typechat
 
 from . import knowledge_schema as kplib
 from ..aitools.embeddings import IEmbeddingModel, NormalizedEmbedding
 from ..storage.memory.semrefindex import collect_action_terms, collect_entity_terms
+from .interfaces import AddMessagesResult
 from .interfaces_core import IKnowledgeExtractor, IMessage, MessageOrdinal, TextLocation
+
+if TYPE_CHECKING:
+    from .conversation_base import ConversationBase
 
 type ChunkOrdinal = int
 
@@ -435,3 +440,76 @@ async def _reassembler_task[TMessage: IMessage](
 
     state.buffered_messages = len(assemblies)
     return state
+
+
+async def add_messages_streaming[TMessage: IMessage](
+    conv: "ConversationBase[TMessage]",
+    messages: AsyncIterable[TMessage],
+    *,
+    batch_size: int = 100,
+    on_batch_committed: Callable[[AddMessagesResult], None] | None = None,
+) -> AddMessagesResult:
+    """Add messages using the pipelined extraction+embedding architecture."""
+    from . import convknowledge
+
+    settings = conv.settings
+    sem_ref_settings = settings.semantic_ref_index_settings
+    storage = await settings.get_storage_provider()
+    knowledge_extractor = (
+        sem_ref_settings.knowledge_extractor or convknowledge.KnowledgeExtractor()
+    )
+    embedding_model = settings.embedding_model
+
+    initial_message_id: MessageOrdinal = await conv.messages.size()
+
+    total = AddMessagesResult()
+
+    def _accumulate(result: AddMessagesResult) -> None:
+        total.messages_added += result.messages_added
+        total.semrefs_added += result.semrefs_added
+        total.chunks_added += result.chunks_added
+        if on_batch_committed:
+            on_batch_committed(result)
+
+    async def _commit_batch(
+        messages_batch: list[TMessage],
+        chunk_results: list[ChunkProcessingResult[TMessage]],
+    ) -> None:
+        result = await conv._commit_batch_from_chunk_results(
+            storage, messages_batch, chunk_results
+        )
+        _accumulate(result)
+
+    chunk_queue: asyncio.Queue[ChunkWorkItem[TMessage] | None] = asyncio.Queue()
+    result_queue: asyncio.Queue[ChunkProcessingResult[TMessage] | None] = asyncio.Queue()
+    stop_state = PipelineStopState()
+    producer_state = ProducerState(next_message_id=initial_message_id)
+
+    async with asyncio.TaskGroup() as tg:
+        tg.create_task(
+            _producer_task(messages, chunk_queue, stop_state, producer_state)
+        )
+        tg.create_task(
+            _dispatcher_task(
+                chunk_queue,
+                result_queue,
+                stop_state,
+                knowledge_extractor,
+                embedding_model,
+                concurrency=sem_ref_settings.concurrency,
+            )
+        )
+        tg.create_task(
+            _reassembler_task(
+                result_queue,
+                stop_state,
+                first_uncommitted_ordinal=initial_message_id,
+                target_commit_chunk_count=batch_size,
+                commit_batch=_commit_batch,
+            )
+        )
+
+    if producer_state.exception is not None:
+        raise producer_state.exception
+
+    return total
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -91,6 +91,58 @@ async def _producer_task[TMessage: IMessage](
             await chunk_queue.put(None)
 
 
+async def _worker_task[TMessage: IMessage](
+    chunk_queue: asyncio.Queue[ChunkWorkItem[TMessage] | None],
+    result_queue: asyncio.Queue[ChunkProcessingResult],
+    stop_state: PipelineStopState,
+    knowledge_extractor: IKnowledgeExtractor,
+    message_embedding_model: IEmbeddingModel,
+    related_terms_embedding_model: IEmbeddingModel | None = None,
+) -> None:
+    """Consume chunk work items and produce chunk processing results.
+
+    Workers stop when they receive a ``None`` sentinel from ``chunk_queue``.
+    Chunks at or beyond ``stop_at_message_id`` are skipped and reported as
+    error results so downstream code can account for them deterministically.
+    """
+    while True:
+        work_item = await chunk_queue.get()
+        if work_item is None:
+            return
+
+        stop_at = stop_state.stop_at_message_id
+
+        if work_item.message_id >= stop_at:
+            await result_queue.put(
+                ChunkProcessingResult(
+                    chunk_id=work_item.chunk_id,
+                    message_id=work_item.message_id,
+                    error=RuntimeError(
+                        "Chunk skipped because stop_at_message_id "
+                        f"is {stop_at} and message_id is {work_item.message_id}"
+                    ),
+                )
+            )
+            continue
+
+        result = await process_chunk_with_extraction_and_embeddings(
+            chunk_id=work_item.chunk_id,
+            message_id=work_item.message_id,
+            chunk_text=work_item.chunk_text,
+            knowledge_extractor=knowledge_extractor,
+            message_embedding_model=message_embedding_model,
+            related_terms_embedding_model=related_terms_embedding_model,
+        )
+
+        if result.error is not None:
+            stop_state.stop_at_message_id = min(
+                stop_state.stop_at_message_id,
+                work_item.message_id,
+            )
+
+        await result_queue.put(result)
+
+
 @dataclass
 class ChunkProcessingResult:
     """Result of processing a single chunk through extraction and embeddings.
--- a/src/typeagent/knowpro/add_messages.py
+++ b/src/typeagent/knowpro/add_messages.py
@@ -134,6 +134,7 @@ async def _dispatcher_task[TMessage: IMessage](
     knowledge_extractor: IKnowledgeExtractor,
     embedding_model: IEmbeddingModel,
     concurrency: int,
+    skip_failed_messages: bool,
 ) -> None:
     """Dispatch chunk work items to bounded per-item worker tasks.
 
@@ -145,6 +146,10 @@ async def _dispatcher_task[TMessage: IMessage](
     tasks run simultaneously.  Chunks at or beyond ``stop_at_message_id`` are
     skipped and reported as error results so the reassembler can account for
     them deterministically.
+
+    Args:
+        skip_failed_messages: If True, don't halt producer on extraction/embedding
+            failures; continue processing. If False, halt on first failure.
     """
     sem = asyncio.Semaphore(concurrency)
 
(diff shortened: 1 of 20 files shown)


gvanrossum:
@KRRT7 If you still care about typeagent-py I'd appreciate your review!

KRRT7:
I'll review it in the morning

tools/ingest_email.py
```diff
@@ -341,15 +352,14 @@ async def _email_generator(
...
-
         # Truncate chunks if --max-chunks is set
+        original_chunk_count: int | None = None
         if max_chunks is not None and len(email.text_chunks) > max_chunks:
-            if verbose:
-                print(f"    Truncating {len(email.text_chunks)} chunks to {max_chunks}")
+            original_chunk_count = len(email.text_chunks)
             email.text_chunks = email.text_chunks[:max_chunks]
+        if verbose:
+            print(label)
```
bmerkle, line 361: for verbose maybe use logging.warning instead of print()

src/typeagent/knowpro/add_messages.py
```diff
@@ -0,0 +1,677 @@
...
+                    chunk_id=work_item.chunk_id,
+                    chunk_text=work_item.chunk_text,
+                    chunk_count=work_item.chunk_count,
+                    message=work_item.message,
+                    knowledge_extractor=knowledge_extractor,
+                    embedding_model=embedding_model,
+                )
+                if result.error is not None:
+                    if skip_failed_messages:
+                        print(
```
bmerkle: for verbose maybe use logging.warning instead of print()

src/typeagent/knowpro/add_messages.py
```diff
@@ -0,0 +1,677 @@
...
+                if skip_failed_messages:
+                    # Skip this failed message and continue
+                    # Find the error from one of the chunks for logging
+                    error_msg = "Unknown error"
+                    for chunk_result in assembly.chunks.values():
+                        if chunk_result.error is not None:
+                            error_msg = str(chunk_result.error)
+                            break
+                    print(
+                        f"Skipping message {state.first_uncommitted_ordinal} "
```
bmerkle, line 433: for verbose maybe use logging.warning instead of print()

tests/test_add_messages_pipeline.py
```diff
@@ -0,0 +1,878 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from __future__ import annotations
```
bmerkle, line 4: AGENTS.md explicitly prohibits this: "from __future__ import annotations is not allowed."

src/typeagent/knowpro/add_messages.py
```diff
@@ -0,0 +1,677 @@
...
+        result.error = e
+        return result
+
+    result.related_terms = _collect_related_terms_for_fuzzy_index(
+        result.extracted_knowledge  # type: ignore[arg-type]
+    )
+
+    # Step 2: Generate embeddings (only if extraction succeeded)
+    try:
+        result.chunk_embedding = await embedding_model.get_embedding_nocache(chunk_text)
```
bmerkle: chunk_embedding is computed but never written to the message text index

src/typeagent/knowpro/add_messages.py
```diff
@@ -0,0 +1,677 @@
...
+@dataclass
+class ReassemblerResult:
+    """Progress and counters produced by the reassembler stage."""
+
+    first_uncommitted_ordinal: MessageOrdinal
+    messages_committed: int = 0
+    chunks_committed: int = 0
+    chunk_failures: int = 0
+    messages_skipped: int = 0
+    buffered_messages: int = 0
```
bmerkle, line 362: buffered_messages is set but never read (only in test code)

tools/ingest_email.py
gvanrossum: We could use logging in a lot of places, but I'd prefer not to switch to that in the current PR. (There's currently only one place where logging is used -- in the stamina callback. Probably I used a different agent for that... :-)

src/typeagent/knowpro/add_messages.py
gvanrossum: Not yet.

tests/test_add_messages_pipeline.py
gvanrossum: Dang. This was written before I put that in AGENTS.md. Will fix.

src/typeagent/knowpro/add_messages.py
gvanrossum: That's fine then.

src/typeagent/knowpro/add_messages.py
gvanrossum: That's a deep one! Need to dig more for this.

tests/test_add_messages_pipeline.py
gvanrossum: Actually I won't fix it in this PR. It's endemic both in tests and src. Let's make an issue of it.

gvanrossum (commented):
@bmerkle It's ready for your re-review. I fixed the chunk embeddings issue. Also had added one more PR since your review came in.

src/typeagent/knowpro/interfaces_storage.py
```diff
@@ -112,6 +112,17 @@ class IMessageCollection[TMessage: IMessage](
 ):
     """A collection of Messages."""
 
+    async def extend(
+        self, items: Iterable[TMessage], chunk_embeddings: list[Any] | None = None
```
bmerkle: instead of list[Any] we could use list[NormalizedEmbedding]

src/typeagent/knowpro/conversation_base.py
```diff
@@ -217,216 +218,169 @@ async def add_messages_streaming(
...
+                knowledge_items,
+            )
 
-            await self._update_secondary_indexes_incremental(start_points)
+            await self._update_secondary_indexes_incremental_with_embeddings(
+                start_points,
+                messages_batch,
+                fuzzy_terms,
+                fuzzy_term_embeddings,
+                chunk_embeddings,
```
bmerkle: parameter chunk_embeddings is not used/dead

src/typeagent/knowpro/add_messages.py
```diff
@@ -0,0 +1,677 @@
...
+            result.error = RuntimeError(
+                f"Knowledge extraction failed: {knowledge_result.message}"
+            )
+            return result
+    except Exception as e:
+        result.error = e
+        return result
+
+    result.related_terms = _collect_related_terms_for_fuzzy_index(
+        result.extracted_knowledge  # type: ignore[arg-type]
```
bmerkle: At this point result.extracted_knowledge is guaranteed non-None by the preceding success check. Prefer assert result.extracted_knowledge is not None (gives a cleaner pyright-clean narrowing and a real runtime check on programmer errors).

src/typeagent/podcasts/podcast.py
```diff
@@ -119,16 +118,9 @@ async def deserialize(
...
         if message_index_data is not None:
             secondary_indexes = self._get_secondary_indexes()
-            # Assert the message index is empty before deserializing
             assert (
                 secondary_indexes.message_index is not None
             ), "Message index should be initialized"
-
-            if isinstance(secondary_indexes.message_index, MessageTextIndex):
-                index_size = await secondary_indexes.message_index.size()
-                assert (
```
bmerkle, line 129: The assertion index_size == 0 before calling message_index.deserialize() was removed from both Podcast.deserialize() and Transcript.deserialize() because MemoryMessageCollection.extend() now auto-populates the message index when messages are loaded. The question is: does MessageTextIndex.deserialize() replace the in-memory embedding store or append to it?

If it replaces (correct), there is no bug. If it appends, deserializing on top of an already-indexed collection would create duplicate entries in TextLocationIndex._text_locations and the underlying embedding array, silently corrupting subsequent similarity searches. Worth adding a comment confirming the replace-not-append semantics, or reinstating a cleared-on-deserialize assertion.

src/typeagent/transcripts/transcript.py
```diff
@@ -120,16 +119,9 @@ async def deserialize(
...
             secondary_indexes = self._get_secondary_indexes()
-            # Assert the message index is empty before deserializing
             assert (
                 secondary_indexes.message_index is not None
             ), "Message index should be initialized"
-
-            if isinstance(secondary_indexes.message_index, MessageTextIndex):
-                index_size = await secondary_indexes.message_index.size()
-                assert (
-                    index_size == 0
```
bmerkle, line 131: The assertion index_size == 0 before calling message_index.deserialize() was removed from both Podcast.deserialize() and Transcript.deserialize() because MemoryMessageCollection.extend() now auto-populates the message index when messages are loaded. The question is: does MessageTextIndex.deserialize() replace the in-memory embedding store or append to it?

If it replaces (correct), there is no bug. If it appends, deserializing on top of an already-indexed collection would create duplicate entries in TextLocationIndex._text_locations and the underlying embedding array, silently corrupting subsequent similarity searches. Worth adding a comment confirming the replace-not-append semantics, or reinstating a cleared-on-deserialize assertion.

gvanrossum:
@bmerkle Claude and I addressed all your feedback. Please re-review.

src/typeagent/knowpro/conversation_base.py
gvanrossum: ~~Fixed by [b2fa6ce](https://github.com/microsoft/typeagent-py/pull/277/commits/b2fa6ce2caa1910b53f1ca5e42a06535fd2df381) (I think).~~

src/typeagent/knowpro/add_messages.py
gvanrossum: Fixed by [efe1ed2](https://github.com/microsoft/typeagent-py/pull/277/commits/efe1ed2c8633fc8a28576d92e1562f9a773769a4).

src/typeagent/podcasts/podcast.py
gvanrossum: Fixed by [b2fa6ce](https://github.com/microsoft/typeagent-py/pull/277/commits/b2fa6ce2caa1910b53f1ca5e42a06535fd2df381). It was a bit tricky.

src/typeagent/transcripts/transcript.py
gvanrossum: Fixed (again) by [b2fa6ce](https://github.com/microsoft/typeagent-py/pull/277/commits/b2fa6ce2caa1910b53f1ca5e42a06535fd2df381).

src/typeagent/knowpro/interfaces_storage.py
gvanrossum: Fixed by [cffb4c3](https://github.com/microsoft/typeagent-py/pull/277/commits/cffb4c3876b13445fe3b723436bd69dfbffdf41f).

src/typeagent/knowpro/conversation_base.py
gvanrossum: Removed by [db13cb0](https://github.com/microsoft/typeagent-py/pull/277/commits/db13cb052c26518d29d24d686369abd24c787789).

KRRT7:
## Review findings

All 734 tests pass on the branch. The prior-round issues (type narrowing, dead param, type tightening, deserialization wasted work) look good. The following are new observations not covered in earlier threads.

---

### Correctness

**`list[typing.Any]` still in SQLite backend** (`sqlite/collections.py:192`)

`cffb4c3` fixed `interfaces_storage.py` and `memory/collections.py` but left the SQLite `extend()` signature as `list[typing.Any]`. The Protocol and the concrete SQLite class are now inconsistent — Pyright won't catch type misuse on the SQLite-backed path.

**Silent `if embedding is not None` in `_commit_batch_from_chunk_results`** (`conversation_base.py:310–312`)

```python
embedding = chunk_embedding_map.get((msg_ord, chunk_ord))
if embedding is not None:
    chunk_embeddings.append(embedding)
```

The validation loop above this already raises `ValueError` for any `None` chunk embedding, so this guard is dead for the non-zero-chunk case. A missing entry in the map means a programmer error (the reassembler staged a message whose chunks weren't all processed), but the silent skip surfaces as a cryptic length-mismatch error one level down in `add_messages_starting_at_with_embeddings`. Prefer an explicit assertion.

**`_commit_if_needed` staged state not cleared before commit** (`add_messages.py:401–409`)

If `commit_batch` raises, `staged_messages` is not cleared. The reassembler's `finally` block calls `_drain_consecutive_complete(force=True)`, which calls `_commit_if_needed(force=True)`, which retries the same staged messages. For the common failure modes (validation error before the transaction, or a DB error that rolls back) this is harmless. But `commit_batch` calls `_accumulate`, which calls the user-provided `on_batch_committed`. If that callback raises *after* a successful DB write, the retry double-commits the same messages. Fix: clear `staged_messages`/`staged_results`/`staged_chunks` before `await commit_batch(...)`, not after.

**Missing comment on replace-not-append semantics** (`podcast.py:81`, `transcript.py:82`)

`b2fa6ce` correctly adds `index_messages=False` so no indexing happens during `extend()`, making the old `index_size == 0` assertion redundant. But the question from the prior review — *does `MessageTextIndex.deserialize()` replace or append?* — is answered by the code (`TextToTextLocationIndex.deserialize` calls `.clear()` then reassigns; `VectorBase.deserialize` does `self._vectors = data`), not by any comment. Worth a one-liner explaining why the assertion is safe to remove.

---

### Performance

**Unnecessary list allocations in hot path** (`add_messages.py:260`)

`_collect_related_terms_for_fuzzy_index` is called for every chunk in every worker:

```python
for action in list(knowledge.actions) + list(knowledge.inverse_actions):
```

This allocates three lists (two coercions + one concatenation) to iterate once. Use `itertools.chain(knowledge.actions, knowledge.inverse_actions)`.

**Extraction and chunk embedding are sequential** (`add_messages.py:307–329`)

`get_embedding_nocache(chunk_text)` runs after extraction completes, but it only needs `chunk_text` — it's independent of the extraction result. Related-term embeddings do depend on extraction, so they must remain after. Using `asyncio.gather` for the first two:

```python
knowledge_result, chunk_embedding = await asyncio.gather(
    knowledge_extractor.extract(chunk_text),
    embedding_model.get_embedding_nocache(chunk_text),
)
```

At `concurrency=10`, each worker saves ~embedding-latency (~20 ms) per chunk off the critical path. On extraction failure the embedding runs unnecessarily, but at low error rates that's acceptable.

**`[e for e in numpy_array]` in three places**

`result.related_term_embeddings = [e for e in rel_embeddings]` (add_messages.py:329), and the same pattern in `memory/reltermsindex.py:295` and `sqlite/reltermsindex.py:219`. All three are just `list(rel_embeddings)` / `list(embeddings)`.

**Semaphore held through `result_queue.put`** (`add_messages.py:196–198`)

```python
    try:
        ...compute result...
        await result_queue.put(result)   # can block when queue full
    finally:
        sem.release()                    # held until put() completes
```

When the reassembler is blocked on a commit, `result_queue` fills to `concurrency * 2`. Workers that have already finished their LLM/embedding work still hold semaphore slots waiting for a queue slot. The dispatcher stalls, the chunk queue fills, the producer stalls. Moving `sem.release()` to before `result_queue.put()` lets the dispatcher start the next chunk immediately after the expensive work finishes. The only risk: on TaskGroup cancellation, a result put-in-progress is silently dropped — but the reassembler task is cancelled too at that point, so it won't be waiting for that result.

gvanrossum:
@KRRT7, I think Claude and I have addressed your concerns in the most recent 8 commits (one per issue).

KRRT7:
* **Redundant Callback Path**: In `_reassembler_task`, the `on_batch_committed` parameter is currently unused because the orchestrator passes `None` and handles the callback via the `_accumulate` wrapper instead. It might be cleaner to remove this parameter from the reassembler if we intend to keep the current orchestration pattern.
* **Logging Verbosity**: When `skip_failed_messages` is enabled, failures are logged in both the `_dispatcher_task` (per chunk) and `_reassembler_task` (per message). For messages with many chunks, this could result in quite a bit of noise in the logs. We might want to consolidate this to only log once at the message level in the reassembler.

Aside from these, the PR looks solid and ready to go.

gvanrossum:
Thanks! Fixed.

KRRT7:
LGTM now

bmerkle:
I have one failing tests, however not really stable fail. output appended below
on a second, third and following runs, test_mcp_server.py ran fine. All other tests passed sucessfully.

should be proceed with this PR and file a new bug ?

```
------------------------------------------------- Captured stderr call -------------------------------------------------
[05/16/26 11:02:28] INFO     Processing request of type            server.py:727
                             ListToolsRequest                                   
                    INFO     Processing request of type            server.py:727
                             CallToolRequest                                    
[05/16/26 11:02:31] INFO     HTTP Request: POST                  _client.py:1740
                             https://xyz.net/opena                
                             i/openai/deployments/text-embedding                
                             -3-small/embeddings?api-version=202                
                             5-01-01-preview "HTTP/1.1 200 OK"                  
                    INFO     HTTP Request: POST                  _client.py:1740
                             https://xyz.net/opena                
                             i/openai/deployments/text-embedding                
                             -3-small/embeddings?api-version=202                
                             5-01-01-preview "HTTP/1.1 200 OK"                  
=============================================== short test summary info ================================================
FAILED tests/test_mcp_server.py::test_mcp_server_query_conversation_slow - AssertionError("{'answer': 'The provided [ANSWER CONTEXT] mentions Kevin Scott briefly but does not include detaile...
============================================ 1 failed, 744 passed in 22.57s ============================================
make: *** [test] Error 1

typeagent-py on  pipeline [$] is 📦 v0.4.0.dev via 🐍 v3.13.12 (typeagent) took 36s 
```

bmerkle:
additional failing test...

```
tests/test_textlocindex.py .........                                                                                                 [ 89%]
tests/test_timestampindex.py .                                                                                                       [ 89%]
tests/test_transcripts.py ......F..........                                                                                          [ 92%]
tests/test_universal_message.py ..                                                                                                   [ 92%]
tests/test_utils.py ...................................                                                                              [ 97%]
tests/test_vectorbase.py ......................                                                                                      [100%]

================================================================= FAILURES =================================================================
________________________________________________ test_transcript_knowledge_extraction_slow _________________________________________________

really_needs_auth = None, embedding_model = <typeagent.aitools.embeddings.CachingEmbeddingModel object at 0x10f5998b0>

    @pytest.mark.asyncio
    async def test_transcript_knowledge_extraction_slow(
        really_needs_auth: None, embedding_model: IEmbeddingModel
    ):
        """
        Test that knowledge extraction works during transcript ingestion.
    
        This test verifies the complete ingestion pipeline:
        1. Parses first 5 messages from Parrot Sketch VTT file
        2. Creates transcript with in-memory storage (fast)
        3. Runs build_index() with auto_extract_knowledge=True
        4. Verifies both mechanical extraction (entities/actions from metadata)
           and LLM extraction (topics from content) work correctly
        """
        # Use in-memory storage for speed
        settings = ConversationSettings(embedding_model)
    
        # Parse first 5 captions from Parrot Sketch
        vtt_file = PARROT_SKETCH_VTT
        if not os.path.exists(vtt_file):
            pytest.skip(f"Test file {vtt_file} not found")
    
        vtt = webvtt.read(vtt_file)
    
        # Create messages from first 5 captions
        messages_list = []
        # vtt is indexable but not iterable
        for i in range(min(len(vtt), 5)):
            caption = vtt[i]
            if not caption.text.strip():
                continue
    
            speaker = getattr(caption, "voice", None)
            text = caption.text.strip()
    
            # Calculate timestamp from WebVTT start time
            offset_seconds = webvtt_timestamp_to_seconds(caption.start)
            timestamp = format_timestamp_utc(UNIX_EPOCH + timedelta(seconds=offset_seconds))
    
            metadata = TranscriptMessageMeta(
                speaker=speaker,
                recipients=[],
            )
            message = TranscriptMessage(
                text_chunks=[text], metadata=metadata, timestamp=timestamp
            )
            messages_list.append(message)
    
        # Create in-memory collections
        msg_coll = MemoryMessageCollection[TranscriptMessage]()
        await msg_coll.extend(messages_list)
    
        _semref_coll = MemorySemanticRefCollection()
        _semref_index = TermToSemanticRefIndex()
    
        # Create transcript with in-memory storage
        transcript = await Transcript.create(
            settings,
            name="Parrot-Test",
            tags=["test", "parrot"],
        )
    
        # Verify we have messages
        assert len(messages_list) >= 3, "Need at least 3 messages for testing"
    
        # Enable knowledge extraction
        settings.semantic_ref_index_settings.auto_extract_knowledge = True
        settings.semantic_ref_index_settings.concurrency = 10
    
        # Add messages with indexing (this should extract knowledge)
>       result = await transcript.add_messages_with_indexing(messages_list)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_transcripts.py:311: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/typeagent/knowpro/conversation_base.py:193: in add_messages_with_indexing
    await self._add_llm_knowledge_incremental(
src/typeagent/knowpro/conversation_base.py:420: in _add_llm_knowledge_incremental
    await semrefindex.add_batch_to_semantic_ref_index_from_list(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

conversation = Transcript(settings=<typeagent.knowpro.convsettings.ConversationSettings object at 0x10f07b6e0>, storage_provider=<typ....knowpro.secindex.ConversationSecondaryIndexes object at 0x11087cc50>, _query_translator=None, _answer_translator=None)
messages = [ConversationMessage(text_chunks=['[ANIMATION] including dancing Botticelli Venus, which links to pet shop: Mr. Pralin...nversationMessageMeta(speaker='Shopkeeper', recipients=[]), tags=[], timestamp='1970-01-01T00:00:17Z', source_id=None)]
batch = [TextLocation(0, 0), TextLocation(1, 0), TextLocation(2, 0), TextLocation(3, 0), TextLocation(4, 0)]
knowledge_extractor = KnowledgeExtractor(model=<typeagent.aitools.model_adapters.PydanticAIChatModel object at 0x10f0b1e60>, max_chars_per_c... merge_action_knowledge=False, translator=<typechat._internal.translator.TypeChatJsonTranslator object at 0x10f2078a0>)
concurrency = 10

    async def add_batch_to_semantic_ref_index_from_list[
        TMessage: IMessage, TTermToSemanticRefIndex: ITermToSemanticRefIndex
    ](
        conversation: IConversation[TMessage, TTermToSemanticRefIndex],
        messages: list[TMessage],
        batch: list[TextLocation],
        knowledge_extractor: IKnowledgeExtractor,
        concurrency: int = 4,
    ) -> None:
        """Extract knowledge from messages and bulk-add to the semantic ref index."""
        if not batch:
            return
        start_ordinal = batch[0].message_ordinal
    
        text_batch = []
        for tl in batch:
            list_index = tl.message_ordinal - start_ordinal
            if list_index < 0 or list_index >= len(messages):
                raise IndexError(
                    f"Message ordinal {tl.message_ordinal} out of range "
                    f"for list starting at {start_ordinal}"
                )
            text_batch.append(messages[list_index].text_chunks[tl.chunk_ordinal].strip())
    
        knowledge_results = await extract_knowledge_from_text_batch(
            knowledge_extractor,
            text_batch,
            concurrency,
        )
        bulk_items: list[tuple[int, int, kplib.KnowledgeResponse]] = []
        for i, knowledge_result in enumerate(knowledge_results):
            if isinstance(knowledge_result, Failure):
>               raise RuntimeError(
                    f"Knowledge extraction failed: {knowledge_result.message:.150}"
                )
E               RuntimeError: Knowledge extraction failed: Several possible issues may have occurred with the given data.
E               
E               Validation path `actions.0.params.0.str` failed for value `{"name": "item_carried", "v

src/typeagent/storage/memory/semrefindex.py:114: RuntimeError
------------------------------------------------------------ Captured log call -------------------------------------------------------------
ERROR    asyncio:base_events.py:1879 Task exception was never retrieved
future: <Task finished name='Task-1602' coro=<AsyncClient.aclose() done, defined at /Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpx/_client.py:1978> exception=RuntimeError('Event loop is closed')>
Traceback (most recent call last):
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpx/_client.py", line 1985, in aclose
    await self._transport.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpx/_transports/default.py", line 406, in aclose
    await self._pool.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpcore/_async/connection_pool.py", line 353, in aclose
    await self._close_connections(closing_connections)
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpcore/_async/connection_pool.py", line 345, in _close_connections
    await connection.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpcore/_async/connection.py", line 173, in aclose
    await self._connection.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpcore/_async/http11.py", line 258, in aclose
    await self._network_stream.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/httpcore/_backends/anyio.py", line 53, in aclose
    await self._stream.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/anyio/streams/tls.py", line 236, in aclose
    await self.transport_stream.aclose()
  File "/Users/merkle/work/microsoft/typeagent-py/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 1344, in aclose
    self._transport.close()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/merkle/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/asyncio/selector_events.py", line 1216, in close
    super().close()
    ~~~~~~~~~~~~~^^
  File "/Users/merkle/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/asyncio/selector_events.py", line 869, in close
    self._loop.call_soon(self._call_connection_lost, None)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/merkle/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/asyncio/base_events.py", line 833, in call_soon
    self._check_closed()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/merkle/.local/share/uv/python/cpython-3.13.12-macos-aarch64-none/lib/python3.13/asyncio/base_events.py", line 556, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
========================================================= short test summary info ==========================================================
FAILED tests/test_transcripts.py::test_transcript_knowledge_extraction_slow - RuntimeError: Knowledge extraction failed: Several possible issues may have occurred with the given data.
====================================================== 1 failed, 744 passed in 20.32s ======================================================
make: *** [test] Error 1

typeagent-py on  pipeline [$] is 📦 v0.4.0.dev via 🐍 v3.13.12 (typeagent) took 21s 
```

gvanrossum:
I think those tests are inherently flaky -- they depend on the LLM service being up, speedy, and returning the right answer. You can file a bug for that and in the meantime merge this PR.

bmerkle (approved):

gvanrossum:
Thanks!

PR #283, opened 2026-06-09
fix: remap chunk results after skipped messages

KRRT7:
## Summary

- Remap source-stream chunk locations to consecutive storage message ordinals when failed messages are skipped.
- Preserve correct chunk embedding and semantic-reference locations for messages committed after a skipped message.
- Add end-to-end coverage for a failed message followed by a successful one.
- Retain regression coverage for graceful shutdown with a full chunk queue.

## Validation

- make format check test
- GitHub CI and CodeQL


KRRT7:
@bmerkle @robgruen if you wouldn't mind reviewing, thanks

tests/test_add_messages_streaming.py
```diff
@@ -200,6 +200,36 @@ async def test_streaming_extraction_failure_stops_at_failing_message() -> None:
         await storage.close()
 
 
+@pytest.mark.asyncio
+async def test_streaming_skips_failed_message_and_commits_next_message() -> None:
```
robgruen, line 212: I would maybe add one more test that includes a repair & reimport to make sure there are no duplicates.

something like this: *fail -> repair -> replay -> replay*

robgruen (approved):

tests/test_add_messages_streaming.py
KRRT7: I've pushed new tests for that, hopefully they're good.

src/typeagent/knowpro/conversation_base.py
```diff
@@ -301,20 +311,45 @@ async def _commit_batch_from_chunk_results(
...
+                raise ValueError("Chunk results exceed staged messages with chunks")
+
+            knowledge_items = [
+                (
+                    source_to_storage_message_id[source_message_id],
+                    chunk_ordinal,
+                    knowledge,
+                )
+                for source_message_id, chunk_ordinal, knowledge in knowledge_items
+            ]
```
bmerkle: Every sibling invariant check in this function (missing embedding, missing knowledge, length mismatches, the two new group-count checks just above) fails with a named `ValueError`. This lookup is the odd one out: if `source_message_id` were ever missing from `source_to_storage_message_id`, this raises a bare `KeyError` with no context pointing back at the chunk-result remap — exactly the kind of regression this PR's own title ("remap chunk results after skipped messages") suggests is worth guarding against explicitly.

```suggestion
            remapped_knowledge_items: list[
                tuple[MessageOrdinal, int, kplib.KnowledgeResponse]
            ] = []
            for source_message_id, chunk_ordinal, knowledge in knowledge_items:
                storage_message_id = source_to_storage_message_id.get(
                    source_message_id
                )
                if storage_message_id is None:
                    raise ValueError(
                        "No storage message id for chunk result: "
                        f"source_message={source_message_id}, chunk={chunk_ordinal}"
                    )
                remapped_knowledge_items.append(
                    (storage_message_id, chunk_ordinal, knowledge)
                )
            knowledge_items = remapped_knowledge_items
```

bmerkle (commented):
## Review summary

Ran a multi-pass review (correctness, test coverage, comment accuracy, error handling) against the diff. The core fix is sound, but the specific bug this PR claims to fix is not actually covered by a test, and there's one error-handling inconsistency worth closing before merge.

### Critical Issues
None. The positional remap in `_commit_batch_from_chunk_results` was traced against the reassembler's invariants (contiguous per-message result grouping in `chunk_results`, monotonic ordinals, zero-chunk-message handling) and is correct given how `add_messages.py`'s producer/reassembler currently stage results.

### Important Issues
1. **The remap this PR fixes has no test verifying it end-to-end.** Every extractor in the test suite returns an empty `KnowledgeResponse`, so `add_knowledge_batch_to_semantic_ref_index` never actually produces a `SemanticRef`, and no test reads `transcript.semantic_refs`/`semantic_ref_index` after a skip. A regression that reverted just the `knowledge_items` remap (lines 344-351) would pass every test currently in this PR. Suggest adding a test with a non-empty `KnowledgeResponse` (e.g. one entity per chunk, named after the chunk text) across a skip, then asserting the resulting semref's `message_ordinal` points at the correct *storage* message, not the source-stream one.
2. **Narrow scenario coverage.** All skip tests cover exactly one skip of a single-chunk message. Worth adding cases for: two consecutive skips, a skip followed by a multi-chunk message, a zero-chunk message adjacent to a skip, and a replay that crosses a batch boundary — these are the shapes most likely to expose an off-by-one in `result_group_index`.
3. **Unguarded dict lookup breaks this function's own error-handling convention** — see inline suggestion below.
4. Positional pairing is inferred from group counts, not verified by identity. `ChunkProcessingResult` already carries a `message: TMessage` field; asserting `result.message is message` while pairing would catch a future reassembler regression that silently swapped two same-sized messages (a scenario the current count-only checks wouldn't catch).

### Suggestions
- Line 341-342: `"Chunk results exceed staged messages with chunks"` carries no numbers (consumed vs. expected group count) — worth adding for debuggability, consistent with the other guards in this function.
- Line 303 comment ("Store embedding by source message/chunk location until commit remaps it") is a little misleading — the map itself is never remapped; it's read with the source key throughout, while a separate list is built in storage order.

### Strengths
- Transaction semantics are clean: every raise in this function happens before any DB-mutating call, inside a `BEGIN IMMEDIATE`/rollback-on-exception transaction — no path leaves partial state.
- Exception propagation to the caller is fail-fast, no swallowing (`asyncio.TaskGroup`).
- `fail_on_text` on `ControlledExtractor` is a well-justified addition for deterministic failure injection under concurrent extraction.
- `_replay()` realistically mirrors the production importer pattern in `tools/ingest_email.py`.

(Note: `add_messages.py`'s `ChunkProcessingResult` docstring still references a nonexistent `success` property — confirmed that file is unchanged by this PR/identical to `main`, so that's pre-existing and out of scope here, but worth a follow-up cleanup.)

src/typeagent/knowpro/conversation_base.py
robgruen: +1 on this...let's get this one in and then we an approve.

PR #294, opened 2026-07-24
Format 0.0 facet values consistently with other floats

LuShadowX:
When indexing a facet whose value is a float, the code formats it with `:g`, but the guard is `isinstance(value, float) and value` — the `and value` skips `0.0` (it's falsy), so a facet value of `0.0` gets stored as `"0.0"` while every other float uses `:g` (e.g. `2.0` -> `"2"`). The comment right above says "If the value is a float, we use .g format", so the extra truthiness check looks unintended.

Dropped `and value` in the two places that do this (`add_facet` and `collect_facet_properties`) so `0.0` is formatted like any other float (`"0"`). Added a small test. Full offline test suite still passes.

### DIFF
--- a/src/typeagent/storage/memory/propindex.py
+++ b/src/typeagent/storage/memory/propindex.py
@@ -45,7 +45,7 @@ async def add_facet(
         value = facet.value
         if value is not None:
             # If the value is a float, we use .g format store it as a string.
-            if isinstance(value, float) and value:
+            if isinstance(value, float):
                 value = f"{value:g}"
             await property_index.add_property(
                 PropertyNames.FacetValue.value,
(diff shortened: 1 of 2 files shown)


LuShadowX:
@microsoft-github-policy-service agree

gvanrossum (commented):
This appears correct. However, I wonder why it was this way -- let me check the original TypeAgent repo.

gvanrossum (approved):
Yeah, TypeAgent had the correct behavior (0.0 -> "0") all along.

LuShadowX:
Thanks for the quick review and merge!

LuShadowX:
Thanks again for the review on this one. If there are other areas of the Python port you'd like a hand with, feel free to ping me.

PR #297, opened 2026-08-01
Update AGENTS.md date/time guidance, bump pyright, minor cleanup

bmerkle:
## Summary
- Document the half-open `[start, stop)` interval convention for date/time range handling in AGENTS.md (agreed in #198)
- Bump pyright dependency to 1.1.411 and widen the `uv_build` requirement to `<0.12.0`
- Minor import reordering in `src/typeagent/knowpro/answers.py`

## Test plan
- [ ] `make check` / `pytest` pass with the updated pyright version

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,7 @@ AGENTS.md. In all cases show what you added to AGENTS.md.
 - Avoid potential import cycles between conversation orchestration and pipeline modules by using neutral payload protocols/arguments instead of importing concrete pipeline result classes across modules.
 - Prefer ordinal type aliases (e.g., `MessageOrdinal`, `ChunkOrdinal`) over raw `int` in pipeline code for readability.
 - When the user asks to "fix the test only", update tests/mocks first and avoid adding production compatibility fallbacks unless explicitly requested.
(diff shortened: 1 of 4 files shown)


gvanrossum:
Can you compare to how this is handled in the TypeScript version? I recall a long, inconclusive argument about this with @umeshma once.

bmerkle:
Hi @gvanrossum

we both had that discussion IIRC, not umeshma. :-) but it was not inconclusive, rather we decided for the python style 👍 

TS uses an inclusive end, documented that way by Umesh himself, and pads date-only stops to 23:59:59.999. Python's timestamp-index contract already diverges to half-open and its tests enforce it,

we discussed this here https://github.com/microsoft/typeagent-py/pull/198/changes/969da137cb915c98929ee11bf3cf5b524023a32f#r2816800611

The resolution of us was a half-open interval [start, stop) — not an inclusive end padded to 23:59:59.999999. That convention is now codified in email_import.py:274-281 (email_matches_date_filter, docstring: "The range is half-open: [start_date, stop_date)").

I also rejected #296 from another user and commented on that in https://github.com/microsoft/typeagent-py/pull/296#issuecomment-5151839037

the motivation for this PR is to avoid this confusion in future, hence i added this rule to AGENTS.md so that we advise and document it for agents, bots and users, that we intentionally go for half open intervals in the python port.

gvanrossum (approved):

gvanrossum:
So there's no need to fix anything? It appeared to me as if #296 had a point.

bmerkle:
> So there's no need to fix anything? It appeared to me as if #296 had a point.

Dang, you are right, thanks.  I apologize to the user in #296...
The half open convention was only implemented in email import but not in the whole framework.
I have fixed this in #309 can you have a look at that if you have time ?

PR #301, opened 2026-08-01
Use module-qualified imports for storage.memory.reltermsindex

bmerkle:
## Summary
- `reltermsindex` (memory variant) exports a mix of classes (`TermToRelatedTermsMap`, `RelatedTermsIndex`, `TermEmbeddingIndex`) and functions (`build_related_terms_index`, `resolve_related_terms`, `dedupe_related_terms`), so per the import style guidelines in #299/AGENTS.md it defaults to module-qualified.
- Converts the 3 `src/typeagent` call sites (`knowpro/secindex.py`, `knowpro/search.py`, `storage/memory/provider.py`) from direct-symbol imports to `from ..storage.memory import reltermsindex` + qualified calls.
- Bonus: in `search.py`, `resolve_related_terms` was both a free function (imported) and a method on `QueryCompiler` — qualifying the import removes that latent name shadowing.
- The sqlite variant (`storage/sqlite/reltermsindex.py`) exports single classes each and is left as direct-symbol import — already compliant, no change.
- Scope: `src/typeagent` only, per #298 — test files keep direct-symbol imports, which is idiomatic pytest style.
- Phase 2 (1/3) of the import-consistency cleanup tracked in #298.

## Test plan
- [x] `make` (format, check on 3.12/3.14, test, build) — 0 pyright errors, 737 passed / 12 skipped, wheel builds

PR #299, opened 2026-08-01
Document Python import style guidelines in AGENTS.md

bmerkle:
## Summary
- Adds a "Python Import Style Guidelines" section to AGENTS.md: default to module-qualified imports, use direct-symbol imports only for classes/exceptions/constants or to avoid severe clutter, and prohibit wildcard imports (with an explicit exception for the sanctioned re-export aggregator pattern, e.g. `interfaces.py`).
- This documents the decision rule discussed in #112 (see [this comment](https://github.com/microsoft/typeagent-py/issues/112#issuecomment-4270060508)) and is Phase 0 of the import-consistency cleanup tracked in #298.
- Adds `ruff` as a dev dependency and a `make ruff` target, scoped narrowly to deterministically enforcing the wildcard-import ban (`F403`/`F405`) via `[tool.ruff.lint]`, addressing @KRRT7's review request that this be pre-commit/CI-linted rather than prose-only. `interfaces.py`'s aggregator pattern is carved out via `[tool.ruff.lint.per-file-ignores]`. Not wired into `make all`/CI yet — staged per the incremental approach from #116, where broader ruff adoption was discussed and deferred; CI wiring is a follow-up.

Rebased onto latest `main` to resolve merge conflicts — the date/time range guideline, pyright bump, and `answers.py` import reorder that were previously part of this branch's diff are already on `main` via #297, so they dropped out as duplicates during the rebase.

## Test plan
- `make ruff` passes cleanly (verified locally).
- `make format` (isort + black) and `uv run isort --check-only` unaffected.

Issue #298, opened 2026-08-01
Standardize import style (module-qualified vs direct-symbol) across the codebase


bmerkle:
## Context

Follow-up to #112, specifically the discussion started in [this comment](https://github.com/microsoft/typeagent-py/issues/112#issuecomment-4229930436) (import/naming inconsistency slows down both humans and LLM-assisted contributors trying to understand the codebase) and the detailed per-module audit in [this comment](https://github.com/microsoft/typeagent-py/issues/112#issuecomment-4270060508) (module-qualified vs. direct-symbol import style is mixed across `knowledge_schema`/kplib, `secindex`, `convknowledge`, `propindex`, `reltermsindex`, `semrefindex`, `textlocindex`).

AGENTS.md now documents the decision rule (added in this effort):

> **Default to Module-Qualified Imports:** Prefer importing whole modules and using qualified calls to prevent namespace pollution, avoid name clashes, and provide immediate context for where functions or objects originate.
> **Use Direct Symbol Imports Cautiously:** Restrict direct imports to specific scenarios: classes/exceptions/constants, avoiding severe repetitive clutter, or standard-library patterns.
> **Prohibit Wildcard Imports.**

This issue tracks applying that rule consistently, based on an AST-based census of actual import sites across `src/typeagent`, `tests/`, and `tools/`.

## Current state (as audited)

| Module | Exports | Target style | Compliance today |
|---|---|---|---|
| `knowledge_schema` (kplib) | 7 classes only | direct-symbol | ❌ split: 20 sites alias it as `kplib` (10 in `src`, 10 in tests/tools), 32 sites already direct |
| `secindex` | 1 class + 2 functions | module-qualified | ✅ already 100% qualified in `src` (6/6); tests use direct-symbol (separate, idiomatic convention) |
| `convknowledge` | 1 class only | either (already consistent) | ✅ already 100% qualified everywhere (4/4) |
| `textlocindex` | 2 classes only | direct-symbol | ✅ already 100% direct everywhere (5/5) |
| `propindex` (memory variant) | 1 enum + 10 functions + 1 class | module-qualified | ❌ opposite of target: 23/25 sites are direct-symbol today |
| `reltermsindex` (memory variant) | 4 classes + 3 functions | module-qualified | ❌ 100% direct-symbol today (14/14) |
[...]

PR #302, opened 2026-08-01
Use module-qualified imports for storage.memory.propindex

bmerkle:
## Summary
- `propindex` (memory variant) exports an enum (`PropertyNames`), ~10 functions, and a class (`PropertyIndex`). Per the import style guidelines in #299/AGENTS.md, low-frequency symbols default to module-qualified:
  - `knowpro/query.py`: `lookup_property_in_property_index`
  - `knowpro/secindex.py`: `build_property_index`
  - `storage/memory/provider.py`: `PropertyIndex`
  - `storage/sqlite/propindex.py`: `make_property_term_text`, `split_property_term_text` (cross-imported from the memory variant)
- `PropertyNames` is deliberately left as a direct-symbol import everywhere it's used (`query.py`, `searchlang.py`, `searchlib.py`) — it's referenced 20+ times across those files, which squarely matches AGENTS.md's "avoiding severe repetitive visual clutter" exception. Qualifying every `PropertyNames.X` reference would add clutter with no disambiguation benefit.
- The sqlite variant's own class (`SqlitePropertyIndex`) exports a single class and is left as direct-symbol import — already compliant, no change.
- Scope: `src/typeagent` only, per #298.
- Phase 2 (2/3) of the import-consistency cleanup tracked in #298. Builds on the same approach as #301 (reltermsindex).

## Test plan
- [x] `make` (format, check on 3.12/3.14, test, build) — 0 pyright errors, 737 passed / 12 skipped, wheel builds

### DIFF
--- a/src/typeagent/knowpro/query.py
+++ b/src/typeagent/knowpro/query.py
@@ -7,8 +7,9 @@ from dataclasses import dataclass, field
 from typing import cast, Literal, Protocol
 
 from ..aitools.embeddings import NormalizedEmbedding
+from ..storage.memory import propindex
 from ..storage.memory.messageindex import IMessageTextEmbeddingIndex
-from ..storage.memory.propindex import lookup_property_in_property_index, PropertyNames
+from ..storage.memory.propindex import PropertyNames
 from .collections import (
     Match,
     MatchAccumulator,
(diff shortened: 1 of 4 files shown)


bmerkle:
@robgruen would you mind taking a look at this one when you get a chance? Thanks!

PR #300, opened 2026-08-01
Replace kplib alias with direct knowledge_schema symbol imports

bmerkle:
## Summary
- `knowledge_schema` exports classes only (`Quantity`, `Quantifier`, `Facet`, `ConcreteEntity`, `ActionParam`, `Action`, `KnowledgeResponse`), so per the import style guidelines in #299/AGENTS.md this is a direct-symbol-import case.
- Removes the `from . import knowledge_schema as kplib` alias (flagged as a smell in [#112](https://github.com/microsoft/typeagent-py/issues/112#issuecomment-4270060508)) everywhere it appears — 10 files in `src/typeagent`, 8 test files, 2 tool scripts — replacing `kplib.X` with a direct `from .knowledge_schema import X, Y, Z` import of the symbols actually used at each call site.
- Mechanical change only; no behavior change.
- Phase 1 of the import-consistency cleanup tracked in #298.

## Test plan
- [x] `make format check test` — 0 pyright errors (Python 3.12 and 3.14), 737 passed / 12 skipped

PR #305, opened 2026-08-01
Use direct-symbol imports for knowpro.interfaces in storage/sqlite

bmerkle:
## Summary
- `knowpro/interfaces.py` is a Protocol/dataclass/type-alias-only aggregator (explicitly re-exporting via `__all__` from `interfaces_core`/`interfaces_indexes`/`interfaces_search`/`interfaces_serialization`/`interfaces_storage` — exempted from AGENTS.md's "no re-export" rule by its own exception clause). Per the import style guidelines in #299/AGENTS.md, a classes-only module like this defaults to direct-symbol import — same reasoning as the `knowledge_schema`/kplib fix in #300.
- `storage/sqlite/` was the lone subpackage using `import interfaces` + qualified access (7/7 files, internally consistent with itself but diverging from the other 41 files in the codebase, which already import symbols directly from `interfaces`).
- Converts all 7: `collections.py`, `messageindex.py`, `propindex.py`, `provider.py`, `reltermsindex.py`, `semrefindex.py`, `timestampindex.py` — ~42 symbol usages total.
- `storage/sqlite/provider.py` had two separate `from ...knowpro.interfaces import ...` lines (one pre-existing direct, one newly added by this change) — isort merged them into a single import block.
- Found via a whole-project import-style scan, tracked in #298.

**Deferred, separate concern** (not part of this PR): a few files (`conversation_base.py`, `add_messages.py`, `storage/sqlite/provider.py`, `podcast_ingest.py`) partially bypass the `interfaces` aggregator, importing some symbols from `.interfaces` and others directly from `.interfaces_core`/`.interfaces_storage` in the same file. That's a different axis of inconsistency (aggregator vs. defining-submodule) — tracked in #298 as a follow-up, not addressed here.

## Test plan
- [x] `make` (format, check on 3.12/3.14, test, build) — 0 pyright errors, 737 passed / 12 skipped, wheel builds

### DIFF
--- a/src/typeagent/storage/sqlite/collections.py
+++ b/src/typeagent/storage/sqlite/collections.py
@@ -8,27 +8,36 @@ import sqlite3
 import typing
 
 from ...aitools.embeddings import NormalizedEmbedding
-from ...knowpro import interfaces, serialization
+from ...knowpro import serialization
+from ...knowpro.interfaces import (
+    IMessage,
+    IMessageCollection,
+    IMessageTextIndex,
+    ISemanticRefCollection,
+    SemanticRef,
+    SemanticRefData,
+    SemanticRefMetadata,
+    TextLocation,
+    TextRange,
+)
 from .schema import ShreddedMessage, ShreddedSemanticRef
 
 
-class SqliteMessageCollection[TMessage: interfaces.IMessage](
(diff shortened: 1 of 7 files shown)


bmerkle:
@robgruen would you mind taking a look at this one when you get a chance? Thanks!

PR #312, opened 2026-08-05
Fix date-only range stop dropping the final day (half-open [start, stop))

LuShadowX:
Replaces #296 and #309. @bmerkle offered on #296 to hand this back to me to land, and I took him up on it; GitHub refused to reopen #296 after the force-push, hence a new PR. The design and the test shape are his — he is credited on the commit.

A `stop_date` with no time compiled to that day's midnight, so "Jan 1 to Jan 5" became `[Jan 1 00:00, Jan 5 00:00)` and everything on Jan 5 was dropped.

- `searchlang.py`: new `exclusive_stop_from_date_time()`; a date-only stop rolls to the next midnight.
- `interfaces_core.py`: `DateRange.end` is now documented and implemented as exclusive (`__contains__` uses `<`), matching both timestamp index backends, which already filtered exclusively.
- `convutils.py`: docstring only, warning that `get_time_range_for_conversation` returns an inclusive-looking bound.

`end is None` still means "unbounded" in `DateRange` but a point query in the indexes. Left alone — separate decision.

`make format check test`: pyright clean on 3.12 and 3.14, **744 passed, 12 skipped** (737 on `main`; the 7 new tests are the only difference). The regression test fails on `main` and passes here.

### DIFF
--- a/src/typeagent/knowpro/convutils.py
+++ b/src/typeagent/knowpro/convutils.py
@@ -35,6 +35,13 @@ async def get_time_range_for_conversation[
 ](
     conversation: IConversation[TMessage, TIndex],
 ) -> DateRange | None:
+    """The conversation's span, for display in a prompt.
+
+    The end is the last message's own timestamp, or None if it has none (the
+    prompt renders that as "now"). It is an inclusive-looking bound, so do NOT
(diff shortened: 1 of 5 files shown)


bmerkle:
sorry, already fixed in #309

Issue #308, opened 2026-08-04
Check compatibility with Python 3.15 rc1


gvanrossum:
3.15 is coming and the first release candidate is out today. we should at least check whether this is supported yet. (I expect typeagent itself to be fine, but we have some dependencies that may lag.)

gvanrossum:
First problem is tiktoken, which doesn't compile. I filed https://github.com/openai/tiktoken/issues/589

gvanrossum:
To be able to build tiktoken from source you need the Rust compiler. I used

```sh
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```

(I also updated the minimal tiktoken version to 0.13.0.)

The next hurdle is PyO3 -- I get this message (among many others):

```text
        error: the configured Python version (3.15) is newer than PyO3's maximum supported version (3.14)
        = help: this package is being built with PyO3 version 0.27.2
        = help: check https://crates.io/crates/pyo3 for the latest PyO3 version available
        = help: updating this package to the latest version of PyO3 may provide compatibility with this Python version
        = help: set PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 to suppress this check and build anyway using the stable ABI
```

gvanrossum:
Using `export PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1` (from the hint in that message) I got PyO3 to build, but now Pydantic-core doesn't build, so I'm giving up.

gvanrossum:
Either uv or Pydantic did *something* (though I don't understand what) and now pydantic-core builds. But you still need that export for PyO3. I filed https://github.com/PyO3/pyo3/issues/6293.

gvanrossum:
Looks like everything now works -- there's a newer Pydantic version out that fixes the 3.15 issues, and the rest is addressed by installing rustc. I'm cleaning up pyproject.html and uv.lock next.

gvanrossum:
For tiktoken, 3.15 wheels are present as of version 0.14.0.
For numpy, 3.15 wheels are present as of version 2.5.2.
On both macOS and Ubuntu Linux, `uv run --python 3.15 pytest` now passes.
There are still some packages being built from source, but they do build:
```txt
Building regex==2026.7.19
Building jiter==0.16.0
Building aiohttp==3.14.3
```

PR #313, opened 2026-08-06
Start testing with Python 3.15 (rc1 only)

gvanrossum:
Addresses #308.

This required some tweaks of the Pydantic version and waiting for the latest pydantic-core release. You must install rustc to be able to build some dependencies for which no 3.15 wheels exist yet. (And you must write 3.15.0rc1 to make uv understand that you want to use a release candidate.

Also, I added 3.15.0rc1 to various places in ci.yml, and updated the uv version there (needed to test with the 3.15 .0c1 binary).

Finally, I reduced some version boundaries that GPT-5.4-mini thought weren't necessary, to give projects linking with typeagent fewer constraints.


.github/workflows/ci.yml
```diff
@@ -132,10 +134,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: ['3.12', '3.13', '3.14']
+        python-version: ['3.12', '3.13', '3.14', 3.15.0rc1']
```
bmerkle: here is a bug: the 3.15.0rc1' is missing the begin quote string

pyproject.toml
```diff
@@ -87,17 +88,17 @@ known_local_folder = ["conftest"]
 dev = [
   "azure-mgmt-authorization>=4.0.0",
   "azure-mgmt-keyvault>=14.0.1",
-  "black>=26.3.1",
-  "coverage[toml]>=7.13.5",
-  "google-api-python-client>=2.194.0",
-  "google-auth-httplib2>=0.3.1",
-  "google-auth-oauthlib>=1.3.1",
-  "isort>=8.0.1",
+  "black>=24.4.2",
```
bmerkle, line 91: is the a reason to downgrad all those packages (black, converage, google-*, isort)
we have been on newer versions already

pyproject.toml
```diff
@@ -87,17 +88,17 @@ known_local_folder = ["conftest"]
...
+  "google-auth-httplib2>=0.2.0",
+  "google-auth-oauthlib>=1.2.1",
+  "isort>=5.13.2",
   "logfire>=4.32.1",  # So 'make check' passes
   "msgraph-sdk>=1.56.0",
   "opentelemetry-instrumentation-httpx>=0.61b0",
-  "pyright>=1.1.411",
-  "pytest>=9.0.3",
-  "pytest-asyncio>=1.3.0",
+  "pyright>=1.1.390",
```
bmerkle, line 100: same question here, why downgrade pyright, pytest ?

bmerkle (changes requested):
small bug in ci.yml and question regarding downgrade packages

bmerkle (approved):

PR #314, opened 2026-08-06
Continue on python 3.15 integration:  Update dependencies, and fix CI matrix, format gate and venv, fix makefile drift

bmerkle:
follow up to #313 and #308

1. Update dependencies to latest and fix the 3.15 classifier

Also replaces the invalid `"Programming Language :: Python :: 3.15.0rc1"` classifier with `3.15`. The rc form is not in PyPI's trove list, so `release-py.yml` would have been rejected on upload.

 2. Stop assigning to `ndarray.shape` in `VectorBase`

numpy 2.5 deprecates in-place shape assignment. 

3. Make the CI python-version matrix and format gate actually work

fixed ci.yml and windows/mac/linux makefile drifts.


gvanrossum:
Hi Bernhard, I really wish you wouldn't rely on AI to produce PR descriptions -- they are too long to read and contain way more words than needed to describe what's going on in the PR from a high level (and for more details we have the diff). The copilot reviews are similarly distracting, but maybe they're useful to you -- I will ignore those to save my sanity.

In #313 I relaxed the minimum versions of dependencies intentionally so that projects using typeagent as a dependency aren't unnecessarily constrained -- a common reason for unsolveable dependencies is that some dependency A requires a high minimum version of dependency B while another dependency C requires a lower version of dependency B. Please roll back that part of your changes.

Makefile
```diff
@@ -14,7 +14,9 @@ format: venv
 .PHONY: check
 check: venv
 	uv run pyright --pythonversion 3.12 src tests tools examples
+	uv run pyright --pythonversion 3.13 src tests tools examples
 	uv run pyright --pythonversion 3.14 src tests tools examples
+	uv run pyright --pythonversion 3.15 src tests tools examples
```
gvanrossum, line 19: Actually I was intentionally running pyright only for the lowest and the highest version -- running it for all versions takes too much time and doesn't add enough diagnostic power.

pyproject.toml
```diff
@@ -24,32 +24,35 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
   "Programming Language :: Python :: 3.14",
-  "Programming Language :: Python :: 3.15.0rc1",
+  "Programming Language :: Python :: 3.15",
   "Programming Language :: Python :: Implementation :: CPython",
   "Typing :: Typed",
 ]
 dependencies = [
```
gvanrossum, line 39: The changes to this section should probably all be rolled back -- I intentionally didn't push these up (explanation in my first overall comment).

**Exception:** the mcp changes are needed (thanks for the catch!).

pyproject.toml
```diff
@@ -24,32 +24,35 @@ classifiers = [
...
-  "numpy>=2.2.6",
-  "openai>=1.81.0",
-  "pydantic>=2.13.4; python_version >= '3.15'",
-  "pydantic>=2.10.0; python_version < '3.15'",
+  # mcp 2.0 dropped mcp.server.fastmcp; src/typeagent/mcp/server.py needs
+  # porting to the new MCPServer API before the cap can be lifted.
+  "mcp[cli]>=1.29.0,<2",
+  "numpy>=2.5.1",
+  "openai>=2.53.0",
+  # Only pydantic-core 2.48+ (pydantic 2.14) ships cp315 wheels.
```
gvanrossum: I built it with pydantic 2.13.0 and somehow it worked. Did it not for you?

pyproject.toml
```diff
@@ -88,17 +91,17 @@ known_local_folder = ["conftest"]
 dev = [
```
gvanrossum, line 103: Updating the versions in this section is fine -- this determines what the dev tools are run with, not what the wheel depends on.

src/typeagent/aitools/vectorbase.py
```diff
@@ -118,7 +118,7 @@ def add_embedding(
         embedding_array = np.asarray(embedding, dtype=np.float32)
         if self._embedding_size == 0:
             self._set_embedding_size(len(embedding_array))
-            self._vectors.shape = (0, self._embedding_size)
+            self._vectors = self._vectors.reshape(0, self._embedding_size)
```
gvanrossum, line 121: Thanks for these catches!

Makefile
```diff
@@ -64,13 +66,17 @@ release: venv
 .PHONY: venv
 venv: .venv
 
-.venv:
+# Re-sync when the dependencies change, not just when .venv/ is missing.
+# The final 'touch' is needed because 'uv sync' leaves .venv/'s own timestamp
+# alone, which would make this target look perpetually out of date.
+.venv: pyproject.toml uv.lock
 	@echo "(If 'uv' fails with 'No such file or directory', try 'make install-uv')"
 	uv sync -q $(FLAGS)
```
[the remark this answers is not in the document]
gvanrossum: Copilot is right that this causes test failures. My recommended solution is simpler than copilot's splitting flags -- just remove `$(FLAGS)` from this line. I don't even recall why I added it.

PS. Always wait for all tests to run, especially when editing infrastructure files like Makefile or ci.yml, before committing.

gvanrossum:
I expect we'll need one more PR after this to fully nail the 3.15 story. That'll be nice since that will be PR #315. :-)

gvanrossum:
More seriously, you've got several test failures with this error:

```txt
  Error: The version '3.15.0rc1' with architecture 'x64' was not found for Ubuntu 24.04.
```

.github/workflows/ci.yml
```diff
@@ -66,14 +66,24 @@ jobs:
...
-          version: "0.12.1"
+          version: "0.12.2"
+
+      # Without this, 'uv sync' downloads a managed Python of its own choosing
+      # and the python-version matrix has no effect at all.
+      - name: Point uv at the matrix interpreter
+        shell: bash
+        env:
+          PYTHON_PATH: ${{ steps.setup-python.outputs.python-path }}
+        run: echo "UV_PYTHON=$PYTHON_PATH" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
bmerkle: ignore for now

Makefile
bmerkle: added as comment and removed 3.13 and 3.14

make.bat
```diff
@@ -25,17 +25,22 @@ if /I "%~1"=="help" goto help
...
 goto help
 
+:: Extra arguments are passed on to the tools, e.g. '.\make format --check'.
 :format
 if not exist ".venv\" call make.bat venv
 echo Formatting code...
-uv run isort src tests tools examples
-uv run black -tpy312 src tests tools examples
+uv run isort src tests tools examples %2 %3 || exit /b 1
+uv run black -tpy312 src tests tools examples %2 %3 || exit /b 1
```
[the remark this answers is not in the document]
bmerkle: fixed

Makefile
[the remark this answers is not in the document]
bmerkle: fixed, thanks

pyproject.toml
bmerkle: 2.13.5 should be ok ?

pyproject.toml
bmerkle: ok, i added manually :-) comments in pyproject.toml about the reason 
added also instructions in AGENTS.md so that it is not updated by the agents (grrrr)

bmerkle:
> More seriously, you've got several test failures with this error:
> 
> ```
>   Error: The version '3.15.0rc1' with architecture 'x64' was not found for Ubuntu 24.04.
> ```

I have no Ubuntu 24.04 to test locally. Do you have ? (I could setup in a VM on my machine but it is some work)

bmerkle:
> I expect we'll need one more PR after this to fully nail the 3.15 story. That'll be nice since that will be PR #315. :-)

Dang, 315 was stolen by dependa bot, grrrr ;-)

pyproject.toml
bmerkle: done. Added comment and also advise in agents.md

bmerkle:
all builds are green, feedback has been implemented
cut the PR description down from the agentic blurb to the essential part (and I will do this in future).

.github/workflows/ci.yml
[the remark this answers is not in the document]
gvanrossum: But what to do about it? It does look like this will just fail on Windows, doesn't it? And there's no bash shell on Windows.

gvanrossum (approved):
Despite the one Windows question I think you can go ahead and merge this! Thanks for taking all my random feedback into account.

PS. You can get an Ubuntu VM for a given PR simply by hitting `.` in your browser when on that PR's GitHub page. This starts one in codespaces. (There are a bunch more steps to set up but that's basically it. I used the smallest VM on offer, 2 cores 8 MB, and it worked fine.)

bmerkle:
@gvanrossum thanks for the tip with . on the GH PR webpage. 

regarding the bash on windows question:

shell: bash — on windows-latest this resolves to Git Bash (C:\Program Files\Git\bin\bash.exe), which is preinstalled on GitHub's Windows images so we are fine :-)

Without this line the step would run in PowerShell, where $PYTHON_PATH and >> behave differently — so the explicit shell: bash is what makes one step definition serve both OSes.

PR #316, opened 2026-08-19
For 3.15, require tiktoken >= 0.14.0 and numpy >= 2.5.2

gvanrossum:
I can now run `uv run --python 3.15 pytest` successfully on macOS. Only three packages are still built from source: regex, aiohttp and jiter. They build quickly. Previously, tiktoken failed to build from source unless you had rustc installed; numpy built successfully but very slowly. So this is major progress. Expect more wheel updates for the others.


gvanrossum:
@bmerkle Could you test this on Windows before approving?

robgruen (approved):

gvanrossum:
Thanks @robgruen! How are things?

robgruen:
GREAT!  Just got back from 2.5 week vacation so feeling very refreshed.  Hopefully your travels were as nice as mine!

bmerkle:
Hi, i tested on a plain new Windows box:

uv run --python 3.15 pytest

      [stderr]
      error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools":
      https://visualstudio.microsoft.com/visual-cpp-build-tools/

question:
should be add the info that we now require clang (Mac/Linux) and cl.exe (Windows) because some wheels have to be build on the machine ?

PS. I am also back from vacation :-)

bmerkle:
PS C:\work\typeagent-py> uv run --python 3.15 pytest
Built jiter==0.16.0  
Built regex==2026.7.19
Built aiohttp==3.14.3

Installed 134 packages in 6.15s

runs also on windows11 with Vstudio 2026 installed

bmerkle (approved):

